PR #19, opened 2016-05-17
Fixing Dollar issue for passwords

mvvsubbu:


Extensions/IISWebAppDeploy/Src/Tasks/IISWebAppMgmt/ManageIISWebApp.ps1
```diff
@@ -62,13 +62,13 @@ function Validate-Inputs
...
 
-function Escape-DoubleQuotes
+function Escape-SpecialChars
 {
     param(
         [string]$str
     )
 
-    return $str.Replace('"', '`"')
+    return $str.Replace('"', '`"').Replace('$', '`$')
```
Ajay-MS: Also ` character affects actual string . Do we need to handle this also ?

Ajay-MS:
Looks fine

PR #20, opened 2016-05-20
Updated version for task and extension. Added change log to readme.

mvvsubbu:


codito:
Looks good. We should think a bit more on how to maintain changelogs (I think it is a nice idea :+1:)

PR #26, opened 2016-05-25
Remove not required string.format calls

pascalberger:


mvvsubbu:
Increase patch number

pascalberger:
@mvvsubbu Done

PR #52, opened 2016-06-23
Update to support Action passed in as either Publish or Script

jhandel:
Allow the user to pick if they want to publish (current only behavior)
or script.. if they choose to script they can then enter a path to store
the script.. This is useful when DBAs would like to know what "happened"
or if during a deployment to a lower environment (QA, UAT, Integration)
they would like to output the script for review prior to running the
production deployment.

### DIFF
--- a/Extensions/IISWebAppDeploy/Src/Tasks/SqlDacpacDeploy/DeployToSqlServer.ps1
+++ b/Extensions/IISWebAppDeploy/Src/Tasks/SqlDacpacDeploy/DeployToSqlServer.ps1
@@ -59,7 +59,9 @@ function GetScriptToRun
         [string]$sqlPassword,
         [string]$connectionString,
         [string]$publishProfile,
-        [string]$additionalArguments
+        [string]$additionalArguments,
+        [string]$action,
+	    [string]$outputPath
     )
 
     $sqlPackageScript = Get-Content .\SqlPackageOnTargetMachines.ps1 | Out-String
@@ -68,7 +70,7 @@ function GetScriptToRun
     $sqlPassword = EscapeSpecialChars -str $sqlPassword
     $additionalArguments = EscapeSpecialChars -str $additionalArguments
 
-    $invokeMain = "ExecuteMain -dacpacFile `"$dacpacFile`" -targetMethod $targetMethod -serverName $serverName -databaseName $databaseName -authscheme $authscheme -sqlUsername `"$sqlUsername`" -sqlPassword `"$sqlPassword`" -connectionString `"$connectionString`" -publishProfile `"$publishProfile`" -additionalArguments `"$additionalArguments`""
+    $invokeMain = "ExecuteMain -dacpacFile `"$dacpacFile`" -targetMethod $targetMethod -serverName $serverName -databaseName $databaseName -authscheme $authscheme -sqlUsername `"$sqlUsername`" -sqlPassword `"$sqlPassword`" -connectionString `"$connectionString`" -publishProfile `"$publishProfile`" -additionalArguments `"$additionalArguments`" -action `"$action`" -outputPath `"$outputPath`""
 
     Write-Verbose "Executing main funnction in SqlPackageOnTargetMachines : $invokeMain"
     $sqlDacpacOnTargetMachinesScript = [string]::Format("{0} {1} ( {2} )", $sqlPackageScript,  [Environment]::NewLine,  $invokeMain)
@@ -93,7 +95,9 @@ function Main
     [string]$connectionString,
     [string]$publishProfile,
     [string]$additionalArguments,
-    [string]$deployInParallel
+    [string]$deployInParallel,
+    [string]$action,
+	[string]$outputPath
     )
 
     Write-Verbose "Entering script DeployToSqlServer.ps1"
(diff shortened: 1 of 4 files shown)


RoopeshNair:
Suggestion can we modify this such that it can generate the script for diagnose-ability as well as publish. 
IMO, generating only script in lieu of publish doesn't add value in an automated workflow. It is more of one time action to get the script.

jhandel:
Your statement is 100% valid if you are only going to deploy a dacpac once ever.. However if your going to be making changes to your schema, indexes, users, functionality, etc and your publishing those via a Dacpac in each release, then a team may want/need to log what the dacpac did in each environment it deployed to. 

In our case its actually a requirement by the Data Services Team as that is how they interpretation of SOX requirements (the solution I automating deployments of is a Ordering system  for a very large national grocer).   

Dacpac's are very state based, there is no way to guarantee what a dacpac will do if you don't run it (to publish or script) on the environment its going to be applied to.  (ok yes backup and restore is an option, but; if your automating the deployment, why we asking devs to  pull a production backup, restore to somewhere they have access to and manually run this just so the output can be stored with the the rest of the deployment results.

In short, anytime a database is going to be changed, it is a reasonable expectation what transformations were executed would be valuable to log... And this is is the only facility to capture them deterministically .

RoopeshNair:
@jhandel 

I’m not undermining the usefulness of logging the transformations. 

Just not in favor of the experience of just logging in lieu of publishing. Which means, I might have to modify my workflow so that the dependent tasks if any should be skipped every time I do a trial run. 

If trial run is what is required, then shouldn’t they be running this on a back-up (pre-prod environments), capture/analyze and then promote to prod?

jhandel:
So pre-analyze is not a requirement (though ideal) the requirement is the asset existing for regulatory reasons.. Because we are dealing with pricing data on a 1+ TB database having the log of what a deployment did exactly is crucial in the eyes of our SOX controls and our DBAs/Data Services team.

That is slightly (though also important to us) different then verifying the deployment prior to release...  Below is the workflow we will be using with this modification.

* Environment Dev -> Refresh from prod (SQL Job), Publish Dacpac, Deploy Suite of applications
* Environment Test -> Refresh from Prod (SQL Job), Script Dacpac, Publish Dacpac (run the same ask twice), Deploy Suite of applications
Approval process happens, with all the right QA, Business and operations approvals
* Environment Production -> Backup prod, script dacpac, publish dacpac, Deploy Suite of applications, 

There are a few other steps in QA and prod as log shipping and replication have to be turned off, then turned back on.. but that gives you a general idea.

if Operations wants to review the transformations that will likely go into prod, they can in QA... However, the key for us is to be able to meet regulatory concerns about being able to show an auditor or the business exactly what was executed on production and when..  pointing to a Dacpac while accurate will create more questions and challenges then it solves in regards to auditors :-)... 

If there is a way to script & publish in the same execution of SQLPackage.exe I am all ears, when I reviewed the documentation it looked as though I could only script or publish..

RoopeshNair:
Will update the thread directly.

Roopesh

RoopeshNair:
@jhandel 

Ok got it, instead of prompting for a file to write-to for "Script" action, we can add this an attachment or logs to the timeline record [task logging commands](https://github.com/Microsoft/vsts-tasks/blob/master/docs/authoring/commands.md)

This should work for hosted agents as well.

jhandel:
attaching as an asset is an interesting approach.. If you like we can cancel this pull request and I'll figure out how to attach it to the log as a logging asset or just read it out to the log per your comments above..

jhandel:
So looking into this approach, its actually going to require changing the guts of the actual WinRM call.. Because the actual sqlpackage.exe is being executed on the target machine and not on the VSTS-Agent there are only a few ways for this suggestion to work..

1) Rewrite InvokeRemoteDeployment.ps1 to return a string (that is the script) and then have the agent write that to the log or save it to a file and attach it.
2) Work out how to copy a file from the target machine to the VSTS-Agent 

We could also write the script to the same folder as the DACPAC as a convention but there is no guarantee that the location is actually shared with the VSTS Agent.

In a "Perfect world" the Task would actually copy the DACPAC from the build assets to the Target of the machine, then winRM would run the DACPAC on that machine (like it does now).. But that's not actually what the script currently does.. Currently it requires the DACPAC already be somewhere the target machine can get to..

RoopeshNair:
@jhandel, sure, appreciate your contributions.

mvvsubbu:
In SqlPackageOnTargetMachines.ps1 we are creating the command to run and executing the same in ExecuteMain function. You can construct command to output the script to console by modifying Get-SqlPackageCmdArgs function and invoke the resulting command before invoking actual deployment step.

jhandel:
I was looking there, however, during the initial analysis I didn't noticed that the deploymentLog wasn't writing (missing the -verbose flag to force it to write)... added that flag to see what was in the DeploymentLog and see everything there.. Going to have to figure out how to pull out the script and attach it as embedding in the log is pretty ugly (functional but ugly).

jhandel:
ok so attaching to the log only works if VSTS is hosting the build... I am not sure it would work at all if its a release at all because of the following error when attempting to upload a file to the log
line of code attempting the attachment

`Write-Host "##vso[artifact.upload containerfolder=RemoteDeploymentResults;artifactname=$machineName_DeploymentLog;]$deploymentDetailsFile"`

Resulting log entry

`##[error]Unable to find container for build: 212`

Looking into this I found that the source of this issue is that VSO Tasks don't supporting writing files to the log for releases tasks yet.. You can follow the issue here : https://github.com/Microsoft/vsts-tasks/issues/1796

This all leads me back to the feeling that actually, asking the user where they want the SQL script to be written to is actually a better experience... 

* Off they have to have a shared location for the DACPAC anyways because the VSTS-Task is not handling publishing the dacpac to the machines during deployment as of yet.. 
* Its clear that this level of WinRM and VSO.Task support isn't quite fully baked.. to support the suggested behavior
*  support these changes at all, we will have to change how InvokeRemoteDeployment.ps1 functions because its only inside that script that we can even get to the DeploymentLog (which is the only way to receive feedback from the ps script being run on the remote host.

We can write the SQL script to the Log directly, however that still requires changes to InvokeRemoteDeployment.ps1.  And when you have 350k (or more) worth of SQL that gets written to the log you loose the ability to review the log for that step without downloading the deployment log as a Zip which isn't very user friendly. It also makes reviews by approvers or auditors very difficult because it has the log time-stamp in front of every line.

Once the VSO task mentioned in the above issue is implemented; migrating to support that might work well.. But we still have this fundamental issue with getting the output out of InvokeRemoteDeployment.ps1...  To do that we still need to change InvokeRemoteDeployment.ps1 from only returning the error from Invoke-RemoteDeployment and instead returns the an array of deploymentResults, or an array of custom objects of the error & deploymentLog.

There is enough impact on files outside of the Dacpac deployment, that I would really prefer the team's feedback on approach before moving forward.

RoopeshNair:
@Josh,<mailto:Josh,>  You’re right, Upload file to log is missing in Release.

Adding @Muthukumaran Kasiviswanathan<mailto:muthuk@microsoft.com> for inputs

RoopeshNair

mvvsubbu:
Thanks for proposing your contribution and it is very unfortunate that we couldn't accept the contribution. We are noting this we will address this in future

PR #62, opened 2016-08-04
authscheme value properly managed

lukka:
The user and password must be passed when SQL Server Authentication is selected (that is when the authscheme picklist value is "sqlServerAuthentication" ). The original code compared to the label "SQL Server Authentication", not to the actual picklist value.


Extensions/IISWebAppDeploy/Src/Tasks/SqlDacpacDeploy/SqlPackageOnTargetMachines.ps1
```diff
@@ -376,7 +376,7 @@ function Get-SqlPackageCmdArgs
             $sqlPkgCmdArgs = [string]::Format('{0} /TargetDatabaseName:"{1}"', $sqlPkgCmdArgs, $databaseName)
         }
 
-        if($authscheme -eq "SQL Server Authentication")
+        if($authscheme -eq "sqlServerAuthentication")
```
mvvsubbu: Modify tests that are getting affected as well. You can verify this by running gulp test

mvvsubbu:
Increase task patch version and fix merge conflicts.

niadak:
We have taken the fix in different PR. Thanks a lot for pointing out the fix.

PR #70, opened 2016-08-22
Fixed directory structure of the Extensions to make packaging work

desattir:
I've changed the directory as per our discussions. I've standardized it, keeping the Src and Tests together inside an extension.

Now packaging of all the extensions, works.


omeshp:
looks good, assuming these are only path changes.

PR #76, opened 2016-08-23
Update AppCmdOnTargetMachines.ps1

jfilburn:
Bug: Given binding already exists for a different website.
function Does-BindingExists: added comma to the end of the binding string which falsely identified a used binding from another site that started with the same string.

### DIFF
--- a/Extensions/IISWebAppDeploy/Src/Tasks/IISWebAppMgmt/AppCmdOnTargetMachines.ps1
+++ b/Extensions/IISWebAppDeploy/Src/Tasks/IISWebAppMgmt/AppCmdOnTargetMachines.ps1
@@ -100,7 +100,7 @@ function Does-BindingExists
     Write-Verbose "Checking binding exists for website (`"$siteName`"). Running command : $command"
 
     $sites = Run-Command -command $command -failOnErr $false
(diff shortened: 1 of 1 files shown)


Extensions/IISWebAppDeploy/Src/Tasks/IISWebAppMgmt/AppCmdOnTargetMachines.ps1
```diff
@@ -100,7 +100,7 @@ function Does-BindingExists
     Write-Verbose "Checking binding exists for website (`"$siteName`"). Running command : $command"
 
     $sites = Run-Command -command $command -failOnErr $false
-    $binding = [string]::Format("{0}/{1}:{2}:{3}", $protocol, $ipAddress, $port, $hostname)
+    $binding = [string]::Format("{0}/{1}:{2}:{3},", $protocol, $ipAddress, $port, $hostname)
```
mvvsubbu: Thanks for your contribution.. Few items that you need to take care

1. Increase patch version in task.json
2. Increase patch extension version in vss-extension.json
3. Update change Unreleased section

jfilburn:
Done

mvvsubbu:
Thanks for fixing all the comments. 

BTW: you can keep updating same PR by doing new commits to the branch on which you opened PR

PR #80, opened 2016-08-23
Update vss-extension.json

jfilburn:


mvvsubbu:
Once I merged a PR, other PRs are not allowing until we take latest from master. Can you take latest from master to patch-3 branch. Remaining two changes also add in this branch only then update the PR so that we don't to merge for other branches to get them into with master

jfilburn:
Done.

PR #128, opened 2016-10-17
Hiyadav patch 1

hiyadav:


Extensions/InputEditor/Src/UIExtensions/scripts/InputEditor/editorExtension.ts
```diff
@@ -10,7 +10,14 @@ class sampleViewModel
 
     public setValueOfParameters(initialconfig){
         if(initialconfig.inputValues[initialconfig.target] != undefined) {
-            this.parameters = ko.observableArray(parse(initialconfig.inputValues[initialconfig.target]));
+            try {
+                this.parameters = ko.observableArray(parse(initialconfig.inputValues[initialconfig.target]));
```
sachinma: what do we return on okay in this case?

PR #138, opened 2016-11-03
Template parameters editor for Azure RG task

desattir:


desattir:
Making this a built-in extension

PR #175, opened 2017-01-12
passing variables are array to script

Ajay-MS:


TaskModules/powershell/TaskModuleSqlUtility/SqlQueryOnTargetMachines.ps1
```diff
@@ -91,6 +91,15 @@ function Execute-SqlQueryDeployment
             }
         }
 
+        # Special handing for argument variable
+        if($spaltArguments.Contains("variable")) {
+            $paramString = $spaltArguments.Item("variable")
+            if($paramString) {
+                $params = $paramString.Split(',')
```
chshrikh: you may need to trim here

chshrikh (approved):

TaskModules/powershell/TaskModuleSqlUtility/SqlQueryOnTargetMachines.ps1
```diff
@@ -106,7 +68,8 @@ function Execute-SqlQueryDeployment
             }
         }
 
-        Invoke-Sqlcmd @spaltArguments
+        Write-Verbose "Invoke-SqlCmd arguments : $spaltArgumentsToLogJson  $additionalArguments"
+        Invoke-Expression "Invoke-SqlCmd @spaltArguments $additionalArguments"
```
kmkumaran: 1) Can you pass additional arguments as named parameter.
2) Are there L0 that needs update?

Ajay-MS: 1. Additional argument string itself contains named parameter, it could not be passed as name parameters.

TaskModules/powershell/TaskModuleSqlUtility/SqlQueryOnTargetMachines.ps1
```diff
@@ -106,7 +68,10 @@ function Execute-SqlQueryDeployment
             }
         }
 
-        Invoke-Sqlcmd @spaltArguments
+        $additionalArguments = $additionalArguments.Replace("`$", "```$")
```
kmkumaran: Is this the same way we handled before splatting?

kmkumaran: Can you add some L0 for this? Can you ensure we have validated this with additional arguments having special characters.

Ajay-MS: Before it escaping was handled for other characters like quotes also and before doing remote.

TaskModules/powershell/TaskModuleSqlUtility/SqlQueryOnTargetMachines.ps1
```diff
@@ -106,7 +68,10 @@ function Execute-SqlQueryDeployment
...
+
+        Write-Verbose "Invoke-SqlCmd arguments : $spaltArgumentsToLogJson  $additionalArguments"
+        Invoke-Expression "Invoke-SqlCmd @spaltArguments $additionalArguments"
```
mvvsubbu: What If additional arguments has -option "Val ue", here I am seeing to invoke expression we are sending -option Val ue which may not be the one we want.

mvvsubbu (changes requested):
Fix comment

mvvsubbu (approved):
Looks good

PR #184, opened 2017-01-30
Adding changes for content extraction of TeamCity artifact type

aydey:


omeshp (changes requested):

Extensions/TeamCity/Src/vss-extension.json
```diff
@@ -201,7 +211,8 @@
           },
           {
             "target": "artifactDetails",
-            "resultTemplate": "{ Name: \"{{version}}\", StreamType: \"zip\", downloadUrl : \"{{endpoint.url}}downloadArtifacts.html?buildTypeId={{definition}}&buildId={{version}}\" }"
```
aydey: We are not really using the "downloadUrl" or the "StreamType" later on. Moreover the template doesn't contain the content of the file. So we changed it. Let me know if any changes need to be made.

omeshp: It gets used as a part of download https://github.com/Microsoft/vsts-agent/blob/master/src/Agent.Worker/Release/Artifacts/CustomArtifact.cs#L91

PR #194, opened 2017-02-07
UNC path physical path fix

mvvsubbu:


Extensions/IISWebAppDeploy/Src/Tasks/IISWebAppMgmt/ManageIISWebApp.ps1
```diff
@@ -33,7 +33,7 @@ function Trim-Inputs([ref]$siteName, [ref]$physicalPath, [ref]$poolName, [ref]$w
     Write-Verbose "Triming inputs for excess spaces, double quotes"
 
     $siteName.Value = $siteName.Value.Trim('"', ' ')
-    $physicalPath.Value = $physicalPath.Value.Trim('"', ' ').Trim('\', ' ')
+    $physicalPath.Value = $physicalPath.Value.Trim('"', ' ').TrimEnd('\')
```
KrishnaAdityaB: can you test this change with the string **" \\path \ "**
I guess, earlier it the space after **path** is removed. Now it is not removed. I am not aware of any scenarios that would be impacted - but calling out the difference between previous implementation

KrishnaAdityaB (approved):

kmkumaran (approved):

Extensions/IISWebAppDeploy/Src/Tasks/IISWebAppMgmt/ManageIISWebApp.ps1
mvvsubbu: I checked that case it works.

PR #207, opened 2017-03-20
Updating utility to add support for authentication

mvvsubbu:


TaskModules/powershell/TaskModuleIISManageUtility/AppCmdOnTargetMachines.ps1
```diff
@@ -731,6 +731,35 @@ function Start-Stop-Recycle-ApplicationPool {
     Invoke-VstsTool -Filename $appCmdPath -Arguments $appCmdArgs -RequireExitCodeZero
 }
 
+function ConfigureWebsiteAuthentication {
+    param (
+        [string]$basicAuthentication,
```
kmkumaran: Should we take one single parameter authentication type and use it? This method supports two types today, but extending to other types like (Forms or ASP.Net impersonation etc) in future might mean we change the function contract.

TaskModules/powershell/TaskModuleIISManageUtility/AppCmdOnTargetMachines.ps1
```diff
@@ -731,6 +731,35 @@ function Start-Stop-Recycle-ApplicationPool {
...
+        [string]$windowsAuthentication,
+        [string]$websiteName
+    )
+
+    Write-Verbose "Configuring website authentication"
+    $appCmdPath, $iisVersion = Get-AppCmdLocation -regKeyPath $AppCmdRegKey
+
+    $anonymousAuthentication = $basicAuthentication -ieq "false" -and $windowsAuthentication -ieq "false"
+
+    $appCmdArgs = [string]::Format('set config "{0}" /section:anonymousAuthentication /enabled:{1} /commit:apphost', $websiteName, $anonymousAuthentication)
```
kmkumaran: The logic here seems, we will turn off the rest of the authentication and only enable the one given. Is turning off the rest required? should the task limit to just enabling/disabling the auth type mentioned and not change anything for remaining auth types?

TaskModules/powershell/TaskModuleIISManageUtility/AppCmdOnTargetMachines.ps1
mvvsubbu: Actually the auth types could be all or any one of them. In this kind scenario is it better to take boolen for each auth type.

TaskModules/powershell/TaskModuleIISManageUtility/AppCmdOnTargetMachines.ps1
mvvsubbu: for enabling basic and windows we need to turn off anonymous auth. https://www.youtube.com/watch?v=fZ2GoVnUh6k

TaskModules/powershell/TaskModuleIISManageUtility/AppCmdOnTargetMachines.ps1
```diff
@@ -731,6 +731,35 @@ function Start-Stop-Recycle-ApplicationPool {
...
+    $command = "`"$appCmdPath`" $appCmdArgs"
+    Write-Verbose "Setting anonymous authentication for website '$websiteName'. Running command $command"
+    Invoke-VstsTool -Filename $appCmdPath -Arguments $appCmdArgs -RequireExitCodeZero
+
+
+    $appCmdArgs = [string]::Format('set config "{0}" /section:basicAuthentication /enabled:{1} /commit:apphost', $websiteName, $basicAuthentication)
```
kmkumaran: You can have machines where the Basic authentication is not enabled in Windows features. How is the error handled in such cases. Do we guide users appropriately here? This comment can apply to other auth types as well.

mvvsubbu: When this auth scheme not enabled, no error comes it ignores. I will update help text to guide the same

kmkumaran (approved):

PR #222, opened 2017-04-28
Show ID of project in selection

### LINKED ISSUE #221: TeamCity extension: Show ID in project dropdown
TeamCity supports sub-projects and therefore there can be multiple sub-projects with identical names. Currently its nearly impossible to select the correct project in this case, since only the name is shown:

The project dropdown should therefore also show the ID and not only the name

pascalberger:
Also show ID in project selection to allow to differentiate between different sub-projects of the same name.

Fixes #221

### DIFF
--- a/Extensions/TeamCity/Src/vss-extension.json
+++ b/Extensions/TeamCity/Src/vss-extension.json
@@ -182,7 +182,7 @@
           {
             "target": "project",
             "dataSourceName": "Projects",
-            "resultTemplate": "{ Value : \"{{id}}\", DisplayValue : \"{{{name}}} ({{id}})\" }"
+            "resultTemplate": "{ Value : \"{{id}}\", DisplayValue : \"{{{name}}} ({{{id}}})\" }"
           },
           {
             "target": "definition",
--- a/Extensions/TeamCity/Src/vss-extension.json
+++ b/Extensions/TeamCity/Src/vss-extension.json
@@ -182,7 +182,7 @@
           {
             "target": "project",
             "dataSourceName": "Projects",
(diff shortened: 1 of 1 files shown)


Extensions/TeamCity/Src/vss-extension.json
```diff
@@ -182,7 +182,7 @@
           {
             "target": "project",
             "dataSourceName": "Projects",
-            "resultTemplate": "{ Value : \"{{id}}\", DisplayValue : \"{{{name}}}\" }"
+            "resultTemplate": "{ Value : \"{{id}}\", DisplayValue : \"{{{name}}} ({{id}})\" }"
```
GitHubSriramB: could you clarify why showing the ID in the display name is useful?

pascalberger: @GitHubSriramB Consider the following structure of projects and subprojects in TeamCity:

```
- ProjectA
  - CI
    - Build
  - Daily
    - Build
- ProjectB
  - CI
    - Build
  - Daily
    - Build
```

With the current implementation the build selection will look like this:

- Build
- Build
- Build
- Build

With the changes of this PR like this:

- Build (ProjectA_CI_Build)
- Build (ProjectA_Daily_Build)
- Build (ProjectB_CI_Build)
- Build (ProjectB_Daily_Build)

Beside the visual aspect, I also wasn't able to select a specific build definition with multiple names, since whatever entry I selected it always has chosen the same.

omeshp: better to use triple quotes {{{id}}}

pascalberger: @omeshp I've changed to triple quotes

PR #223, opened 2017-04-28
Show builds from all branches and not only from default branch

### LINKED ISSUE #220: TeamCity extension: Support artifacts from non default branch
The TeamCity extension currently doesn't support artifacts from non default branches. It should be possible to deploy from any branch.

pascalberger:
Currently the TeamCity extension only allows to deploy artifacts from the default branch. This change will show builds from all branches.

Fixes #220

### DIFF
--- a/Extensions/TeamCity/Src/vss-extension.json
+++ b/Extensions/TeamCity/Src/vss-extension.json
@@ -78,7 +78,7 @@
           },
           {
             "name": "LatestBuild",
-            "endpointUrl": "{{endpoint.url}}httpAuth/app/rest/buildTypes/{{definition}}/builds?count=1&status=success",
+            "endpointUrl": "{{endpoint.url}}httpAuth/app/rest/buildTypes/{{definition}}/builds?locator=branch:default:any&count=1&status=success",
             "resultSelector": "jsonpath:$.build[*]"
           },
           {
--- a/Extensions/TeamCity/Src/vss-extension.json
+++ b/Extensions/TeamCity/Src/vss-extension.json
@@ -73,7 +73,7 @@
           },
           {
             "name": "Builds",
-            "endpointUrl": "{{endpoint.url}}httpAuth/app/rest/buildTypes/{{definition}}/builds",
+            "endpointUrl": "{{endpoint.url}}httpAuth/app/rest/buildTypes/{{definition}}/builds/?locator=branch:default:any",
             "resultSelector": "jsonpath:$.build[?(@.status=='SUCCESS')]"
           },
           {


Extensions/TeamCity/Src/vss-extension.json
```diff
@@ -73,7 +73,7 @@
           },
           {
             "name": "Builds",
-            "endpointUrl": "{{endpoint.url}}httpAuth/app/rest/buildTypes/{{definition}}/builds",
+            "endpointUrl": "{{endpoint.url}}httpAuth/app/rest/buildTypes/{{definition}}/builds/?locator=branch:default:any",
```
GitHubSriramB: should we take the default branch to use for querying builds as input when linking the TeamCity artifact rather than listing build across all branches?

pascalberger: @GitHubSriramB Sorry, I don't understand what you mean. While linking to an artifact from this extension no builds are listed on the "Link an artifact source" (There's no default version or similar). And to have builds from all branches available in the "Create new release" dialog was the goal of this PR.

omeshp: Strange that the API lists only builds corresponding to default branch without any locator.
@GitHubSriramB I think default version can be added incrementally to TeamCity extension, this change looks ok to list all the builds for now.

omeshp: Do we need similar change for LatestBuild API?

pascalberger: @omeshp @GitHubSriramB This approach obviously doesn't scale very well with a lot of branches or builds. Better would be a solution where the branches can be defined while creating the artifact link, but I don't know if this is posisble from TFS side, and also by the TeamCity REST API (would require wildcards for branches).

pascalberger: @omeshp Whats the expected outcome of `LatestBuild`? Or in which cases is it called? As it is now it only returns builds from the default branch. Depending on its use it would also require a locator to also return builds from other branches.

omeshp: LatestBuild is used for autotrigger scenarios, if you have linked both vsts build artifact and teamcity artifact and setup a trigger on vsts build artifact, then when a build is queued LatestBuild will be used to select the build for teamcity artifact. In short we should do the locator changes for LatestBuild also. For filtering artifact based on branches we should have default version selector but that is another feature.

pascalberger: @omeshp OK, I've also added the locator to the LatestBuild endpoint. But IMHO the extension as it is now is not usable in an autotrigger scenario with multiple artifacts, as it will choose a build from a random branch.

omeshp: @pascalberger I am not sure about what default branch depicts in teamcity configurations. Are you saying it is better to keep the existing behavior of picking up latest build from default branch or do you mean that we should have an option of default version at the time of linking for the extension to be usable in autotrigger scenario?

pascalberger: @omeshp IMHO best solution would be the possibility to define the **branch** used in autotrigger scenario at the time of linking. Until then I think always returning the default branch for LatestBuild is the best option, even though it's a limitation it at least leads to a predictable result.

pascalberger:
@omeshp @GitHubSriramB Do you have any open issues or required changes with this PR?

AntonAnikeev:
Hey, guys. 
Have you released this fix to the latest version of extension? Seems like no.
How can I update my current version of extension? Is it possible?

GitHubSriramB:
@AntonAnikeev - Yes, we have updated TeamCity extension to include this fix.

The latest extension would be used in accounts where it is already installed. Isn't this what you are seeing?

AntonAnikeev:
@GitHubSriramB Great! Thanks for quick answer.

PR #229, opened 2017-06-23
BugFix1018698: Can get available version of external tfs build if it …

omeshp:
…has a double quote in it.


GitHubSriramB:
don't you have to update the extension version?

GitHubSriramB (approved):
don't you have to update extension version?

PR #230, opened 2017-06-23
Fixing node dependencies for no caching mechanism

hililh:


package-utils.js
```diff
@@ -0,0 +1,81 @@
+var fs = require('fs');
+var path = require('path');
+var shell = require('shelljs');
+var ncp = require('child_process');
+
+//------------------------------------------------------------------------------
+// shell functions
+//------------------------------------------------------------------------------
+var rp = function (relPath) {
+    return path.join(shell.pwd() + '', relPath);
```
sachinma: why are you adding the empty string

package-utils.js
```diff
@@ -0,0 +1,81 @@
...
+var shell = require('shelljs');
+var ncp = require('child_process');
+
+//------------------------------------------------------------------------------
+// shell functions
+//------------------------------------------------------------------------------
+var rp = function (relPath) {
+    return path.join(shell.pwd() + '', relPath);
+}
+exports.rp = rp;
```
sachinma: why does it need exporting

sachinma: Have you taken these functions from vsts-tasks or other similar repo? Then it would be good to add that in comment

sachinma: Look if other functions need to be exported either

sachinma (approved):

package-utils.js
hililh: Yes I have taken this functions from vsts-task repo. will add comment.
I am exporting it, because in future we may need this functions in other modules

PR #237, opened 2017-07-18
1 )Ansible Improvements to provide the first class improvement to CLI…

hililh:
1) Ansible Improvements to provide the first class improvement to CLI option. 
2) Adding L0 test for ansible


Extensions/Ansible/Src/Tasks/Ansible/task.json
```diff
@@ -56,69 +60,175 @@
...
+            "defaultValue": "agentMachine",
             "options": {
-                "commands": "Commands",
-                "script": "Shell script"
+                "agentMachine": "Agent machine",
+                "ansibleMachine": "Ansible machine"
             },
-            "helpMarkDown": "Choose to either run shell commands or a shell script on the remote machine."
+            "required": true,
+            "helpMarkDown": "Select agent machine, if playbook is located on automation agent. In this case playbook is copied to ansible machine. Otherwise, select ansible machine.",
```
sachinma: Lets have help text more verbose. I think we should fell free to use multiple lines to explain each option

Extensions/Ansible/Src/Tasks/Ansible/task.json
```diff
@@ -56,69 +60,175 @@
...
+            "name": "playbookPathAnsibleMachine",
             "type": "string",
-            "label": "Arguments",
+            "label": "Path",
             "defaultValue": "",
-            "required": false,
-            "visibleRule": "cliRunOptions = script",
-            "helpMarkDown": "Arguments to pass to the shell script."
+            "required": true,
+            "helpMarkDown": "Path of the playbook files on execute on the ansible machine.",
```
sachinma: path of the playbook file  to execute

Extensions/Ansible/Src/vss-extension.json
```diff
@@ -2,10 +2,10 @@
   "manifestVersion": 1.0,
   "id": "vss-services-ansible",
   "name": "Ansible",
-  "publisher": "ms-devlabs",
+  "publisher": "ms-vscs-rm",
```
sachinma: why this change

Extensions/Ansible/Src/Tasks/Ansible/ansibleCommandLineInterface.ts
```diff
@@ -0,0 +1,201 @@
...
+
+            if (sshClientConnection) {
+
+                if(playbookSource == 'agentMachine') {
+                    
+                    let playbookRoot: string = tl.getInput('playbookRoot', true);
+                    let playbook: string = tl.getInput('playbookPathAgentMachine', true);
+                    
+                    let remotePlaybookRoot = '/tmp/' + path.basename(playbookRoot);
+                    let remotePlaybookRootPath = '"' + remotePlaybookRoot + '"';
```
sachinma: why do you need to put it in quotes

Extensions/Ansible/Src/Tasks/Ansible/ansibleCommandLineInterface.ts
```diff
@@ -0,0 +1,201 @@
...
+                    tl.debug('ansiblePlaybookRootPath = ' + remotePlaybookRootPath);
+
+                    let scpConfig = sshConfig;
+                    scpConfig.path = remotePlaybookRoot;
+                    tl.debug('Copying playbook to ansible machine.');
+                    await ansibleUtils.copyFileToRemoteMachine(playbookRoot, scpConfig);
+
+                    let remotePlaybookPath = remotePlaybookRoot + "/" + path.basename(playbook);
```
sachinma: Is there no path.join like function to join the paths

Extensions/Ansible/Src/Tasks/Ansible/ansibleCommandLineInterface.ts
```diff
@@ -0,0 +1,201 @@
...
+
+                    remoteCommand = remoteCommand.concat(" " + remotePlaybookPath)
+
+                    cleanupCmd.push('rm -rf ' + remotePlaybookRoot);
```
sachinma: are you doing this before copy as well. I think that might be good to do in addition to this

Extensions/Ansible/Src/Tasks/Ansible/ansibleCommandLineInterface.ts
```diff
@@ -0,0 +1,201 @@
...
+
+                        remoteCommand = remoteCommand.concat(' -i ' + remoteInventory);
+
+                        cleanupCmd.push('rm -f ' + remoteInventory);
+
+                    } else if(inventoryFileSource == 'ansibleMachine') {
+
+                        let inventoryFile: string = tl.getInput('inventoryFileAnsibleMachine', true);
+                        tl.debug('InventoryFile = ' + inventoryFile);
+                        remoteCommand = remoteCommand.concat(' -i ' + inventoryFile);
```
sachinma: consider moving this out of both if and else clause and both if and else only set inventoryFile

Extensions/Ansible/Src/Tasks/Ansible/ansibleCommandLineInterface.ts
```diff
@@ -0,0 +1,201 @@
...
+                    tl.debug('RemoteInventoryPath = ' + remoteInventoryPath);
+
+                    let inventoryCmd: string = 'echo ' + '"' + content + '"' + ' > ' + remoteInventory;
+                    await ansibleUtils.runCommandOnRemoteMachine(inventoryCmd, sshClientConnection);
+
+                    let dynamicInventory: boolean = tl.getBoolInput('inventoryInlineDynamic', true);
+                    
+                    if(dynamicInventory == true) {
+                        await ansibleUtils.runCommandOnRemoteMachine('chmod +x ' + remoteInventory, sshClientConnection);
+                    }
```
sachinma: dont we need to add a flag for remote inventory

Extensions/Ansible/Src/Tasks/Ansible/ansibleCommandLineInterface.ts
hililh: To highlight the debug text. This path I am printing for debug

Extensions/Ansible/Src/Tasks/Ansible/ansibleCommandLineInterface.ts
hililh: I am not doing this before copy because it will copy will override, if the file is already present. This is as good as removing the old and pasting the new one. So from my understanding running the clean-up once after the copy should be sufficient

Extensions/Ansible/Src/Tasks/Ansible/ansibleCommandLineInterface.ts
hililh: ok

Extensions/Ansible/Src/vss-extension.json
hililh: Usha want it to publish as a Microsoft and not as MS devlabs

Extensions/Ansible/Src/Tasks/Ansible/ansibleCommandLineInterface.ts
hililh: We cannot use path.join()
agent could be running on windows machine, in this case path would join using back slash. where as this command needs to be run on Linux which require forward slash.
If the agent is on Linux machine then we could have used path.join

Extensions/Ansible/Src/Tasks/Ansible/ansibleCommandLineInterface.ts
hililh: doesn't required

Extensions/Ansible/Src/Tasks/Ansible/task.json
hililh: I feel all this scenario should be captured on Readme file in verbose mode.
Will add it there

Extensions/Ansible/Src/Tasks/Ansible/task.json
hililh: done

Extensions/Ansible/Src/Tasks/Ansible/ansibleCommandLineInterface.ts
sachinma: Consider not  creating var for it in that case

Extensions/Ansible/Src/Tasks/Ansible/ansibleCommandLineInterface.ts
sachinma: makes sense

Extensions/Ansible/Src/Tasks/Ansible/ansibleCommandLineInterface.ts
sachinma: But what if there are extra files

sachinma (approved):

PR #265, opened 2017-09-05
removing azureblobstorageprovider from itemleveldownloader

hiyadav:
As discussed, we shall be moving AzureBlobStorageProvider to Vsts-tasks/Tasks/Common.
This PR is to:
1. Remove AzureBlobStorageProvider from the providers for Item-Level-Downloader and thus removing Azure-Storage as a dependency.
2. Updating patch version of Item-Level-Downloader.


ItemLevelDownloader/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "item-level-downloader",
-  "version": "1.0.10",
+  "version": "1.0.11",
```
omeshp: update package version before checkin.

ItemLevelDownloader/package.json
```diff
@@ -9,8 +9,7 @@
   },
   "dependencies": {
     "handlebars": "4.0.10",
-    "minimatch": "3.0.2",
-    "azure-storage": "2.2.1"
+    "minimatch": "3.0.2"
```
bishal-pdMSFT: Please move shrinkwrap file as well. I had created this file for azure-storage only. Hence you can move it completely. @omeshp have you changed that file? If so, then it please retain your changes. I guess you will need another list once Sriram gets OSS approval.

hiyadav: As shrinkWrap is not present in this Repo. i shall use shrink wrap in vsts-tasks.

ItemLevelDownloader/package.json
hiyadav: i have updated it here to 1.0.11, is there some other version also that needs to be changed ?

bishal-pdMSFT:
Remove azureUpload function from task.ts and related config from config.json.example

bishal-pdMSFT:
Looks good to me

PR #282, opened 2017-09-29
bug fixes plus some tests

omeshp:


Extensions/ArtifactEngine/Providers/filesystemProvider.ts
```diff
@@ -33,21 +33,26 @@ export class FilesystemProvider implements models.IArtifactProvider {
...
             this.ensureDirectoryExistence(folder);
 
-            Logger.logMessage('Downloading '+ item.path + ' to '+ outputFilename);
+            Logger.logMessage('Downloading ' + item.path + ' to ' + outputFilename);
             const outputStream = fs.createWriteStream(outputFilename);
             stream.pipe(outputStream);
             stream.on("end",
                 () => {
                     Logger.logMessage(`Downloaded '${item.path}' to '${outputFilename}'`);
+                    if (!item.metadata) {
```
GitHubSriramB: when will this happen?

Extensions/ArtifactEngine/Providers/webProvider.ts
```diff
@@ -37,12 +38,27 @@ export class WebProvider implements models.IArtifactProvider {
...
+        var promise = new Promise<NodeJS.ReadableStream>(async (resolve, reject) => {
+            if (!artifactItem.metadata || !artifactItem.metadata['downloadUrl']) {
+                reject("No downloadUrl available to download the item.");
+            }
+
             var itemUrl: string = artifactItem.metadata['downloadUrl'];
             itemUrl = itemUrl.replace(/([^:]\/)\/+/g, "$1");
-            let res: httpm.HttpClientResponse = await this.httpc.get(itemUrl);
-            resolve(res.message);
+            var promise = this.httpc.get(itemUrl);
```
GitHubSriramB: don't we need to add any charset utf8 headers in the get call?

Extensions/ArtifactEngine/Engine/artifactEngine.ts
```diff
@@ -96,4 +96,9 @@ export class ArtifactEngine {
 
     private artifactItemStore: ArtifactItemStore = new ArtifactItemStore();
     private logger: Logger;
-}
\ No newline at end of file
+}
+
+process.on('unhandledRejection', (reason, promise) => {
```
GitHubSriramB: what is this fix for?

Extensions/ArtifactEngine/EngineTests/artifactEngineTests.ts
```diff
@@ -10,7 +11,7 @@ mockery.enable({
 mockery.registerMock('fs', {
     createWriteStream: (a) => {
         var mockedStream = stream.Writable();
-        mockedStream._write = () => {};
+        mockedStream._write = () => { };
```
GitHubSriramB: does the additional space matter?

Extensions/ArtifactEngine/Providers/filesystemProvider.ts
```diff
@@ -19,8 +19,8 @@ export class FilesystemProvider implements models.IArtifactProvider {
         return this.getItems(itemsPath, artifactItem.path);
     }
 
-    getArtifactItem(artifactItem: models.ArtifactItem): Promise<Stream.Readable> {
-        var promise = new Promise<Stream.Readable>(async (resolve, reject) => {
+    getArtifactItem(artifactItem: models.ArtifactItem): Promise<NodeJS.ReadableStream> {
```
GitHubSriramB: what is this change for?

Extensions/ArtifactEngine/Store/artifactItemStore.ts
```diff
@@ -30,7 +30,7 @@ export class ArtifactItemStore {
 
     public itemsPendingProcessing(): boolean {
         var processingTickets = this._downloadTickets.filter(x => x.state === models.TicketState.Processing || x.state === models.TicketState.InQueue);
-        return processingTickets ? processingTickets.length != 0 : false;
+        return processingTickets.length != 0;
```
GitHubSriramB: wasn't the previous code safer?

Extensions/ArtifactEngine/task.ts
```diff
@@ -75,6 +77,25 @@ async function downloadVSTSDropWithMultipleFiles(processorOptions) {
     await processor.processItems(webProvider, localFileProvider, processorOptions);
 }
 
+async function downloadVSTSDropWithMultipleFiles2(processorOptions) {
```
GitHubSriramB: what is the difference in this?

Extensions/ArtifactEngine/package.json
```diff
@@ -9,19 +9,21 @@
   },
```
GitHubSriramB: looks like i need to again register for the components we are using now :( are you sure these changes are required?

Extensions/ArtifactEngine/EngineTests/artifactEngineTests.ts
```diff
@@ -51,26 +52,18 @@ describe('artifactEngine.processItems', () => {
     it('should call getArtifactItem only for artifact that match the download pattern', async () => {
         var testProvider = new providers.StubProvider();
         var downloadOptions = new engine.ArtifactEngineOptions();
-        downloadOptions.itemPattern = '*(PAth4|path5)\\**';
+        downloadOptions.itemPattern = '@(PAth4|path5)\\**';
```
GitHubSriramB: is this to validate the scenario of service passing multiple items to download for partial artifacts (the discussion we had with Lova)?

GitHubSriramB (approved):

Extensions/ArtifactEngine/Providers/webProvider.ts
```diff
@@ -37,12 +38,27 @@ export class WebProvider implements models.IArtifactProvider {
...
+            promise.catch(reason => {
+                reject(reason);
+            });
+
+            let res: httpm.HttpClientResponse = await promise;
+
+            if (res.message.headers['content-encoding'] === 'gzip') {
```
GitHubSriramB: for build artifact, were we always getting gzip or is it conditional? wondering how we didn't find this issue so far?

Extensions/ArtifactEngine/Engine/artifactEngine.ts
omeshp: if the code fails at some point in node like de-referencing undefined object it generates an unhandledRejection warning, we need to map it to error to fail in such cases.

Extensions/ArtifactEngine/EngineTests/artifactEngineTests.ts
omeshp: no, but when i autoformat in vscode it generates this space.

Extensions/ArtifactEngine/EngineTests/artifactEngineTests.ts
omeshp: No, it is just to test single patterns * will match Path4Path4 also.

Extensions/ArtifactEngine/Providers/filesystemProvider.ts
omeshp: After applying unzip transformation what we get is a ReadableStream which is the base call for Stream.Readable, we need to change the signatures to accommodate for that.

Extensions/ArtifactEngine/Providers/filesystemProvider.ts
omeshp: depened on the template currently all templates have metadata, if there is no metadata we should explicitly reject here.

Extensions/ArtifactEngine/Providers/webProvider.ts
omeshp: not required, for the testing I did. Will take another look to check if it is required or not later.

Extensions/ArtifactEngine/Providers/webProvider.ts
omeshp: It is conditional, I think it depends on size and api version but need to confirm.

Extensions/ArtifactEngine/Store/artifactItemStore.ts
omeshp: The code was dead code, filter never returns undefined. It was caught using codecoverage.

GitHubSriramB (approved):

PR #311, opened 2017-11-21
add telemetry

omeshp:


Extensions/ArtifactEngine/Engine/cilogger.ts
```diff
@@ -0,0 +1,21 @@
...
+        releaseuri: tl.getVariable('Release.ReleaseUri'),
+        releaseid: tl.getVariable('Release.ReleaseId'),
+        builduri: tl.getVariable('Build.BuildUri'),
+        buildid: tl.getVariable('Build.Buildid')
+    };
+}
+
+export function publishEvent(feature, properties: { [key: string]: any }): void {
+    try {
+        tl.assertAgent('2.125.0');
```
GitHubSriramB: what happens when agent is less than 2.125.0? will this not log telemetry? or will it fail?

Extensions/ArtifactEngine/Engine/cilogger.ts
```diff
@@ -0,0 +1,21 @@
...
+        tl.publishTelemetry(area, feature, Object.assign(getDefaultProps(), properties));
+    } catch (err) {
+        tl.debug('Unable to publish telemetry due to lower agent version.');
```
GitHubSriramB: since publishTelemetry is also within try block, suggest keeping the log message generic:

Unable to publish telemetry, error: {err}. I assume the error message will include detail of agent version if tl.assertAgent() throws error. if not, let's log agent version also along with the message, like:

Unable to publish telemetry, error: {err}, agentVersion: {agentVersion}

Extensions/ArtifactEngine/Engine/cilogger.ts
```diff
@@ -0,0 +1,21 @@
...
+        releaseid: tl.getVariable('Release.ReleaseId'),
+        builduri: tl.getVariable('Build.BuildUri'),
+        buildid: tl.getVariable('Build.Buildid')
+    };
+}
+
+export function publishEvent(feature, properties: { [key: string]: any }): void {
+    try {
+        tl.assertAgent('2.125.0');
+        tl.publishTelemetry(area, feature, Object.assign(getDefaultProps(), properties));
```
GitHubSriramB: if download fails on lower agent versions, it will hinder us from going forward. we need telemetry to know what issues are getting it. do you suggest we should run the config change again? or should we log telemetry using the older way?

Extensions/ArtifactEngine/Store/artifactItemStore.ts
```diff
@@ -55,7 +55,17 @@ export class ArtifactItemStore {
...
 
-    public increaseRetryCount(item: models.ArtifactItem){
+    public getRootLocation(): string {
+        var rootItem = this._downloadTickets.find(x => x.artifactItem.path === "");
+        var rootLocation = '';
+        if (rootItem && rootItem.artifactItem.metadata) {
+            rootLocation = rootItem.artifactItem.metadata["downloadUrl"];
+        }
+
+        return rootLocation ? rootLocation : '';
```
GitHubSriramB: when can this be ''?

Extensions/ArtifactEngine/StoreTests/artifactItemStoreTests.ts
```diff
@@ -276,4 +276,40 @@ describe('artifactItemStore.increaseRetryCount', () => {
...
+        artifactItem1.path = "path1";
+
+        artifactItemStore.addItem(artifactItem1);
+
+        assert.equal(artifactItemStore.getRootLocation(), '');
+    });
+});
+
+describe('artifactItemStore.getRootLocation', () => {
+    it('should return if root item does not metadata', () => {
```
GitHubSriramB: ... does not **have** metadata

PR #312, opened 2017-11-24
pattern match fixes

gykuma:
Update logic for including/excluding patterns
Resolve comments of [previous PR](https://github.com/Microsoft/vsts-rm-extensions/pull/309)

### DIFF
--- a/Extensions/ArtifactEngine/Engine/artifactEngine.ts
+++ b/Extensions/ArtifactEngine/Engine/artifactEngine.ts
@@ -118,27 +118,32 @@ export class ArtifactEngine {
         }
     }
 
-    shouldDownload(path: string) {
-        var shouldDownload = false;
+    shouldDownload(path: string): boolean {
+        let matchedPattern;
         for (let pattern of this.patternList) {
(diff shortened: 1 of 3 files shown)


Extensions/ArtifactEngine/Engine/artifactEngine.ts
```diff
@@ -118,27 +118,32 @@ export class ArtifactEngine {
...
                 }
             }
         }
-        return shouldDownload;
+
+        return (!matchedPattern || matchedPattern.startsWith('-')) ? false : true;
     }
 
+
+    //Supported pattern strings:
```
GitHubSriramB: we should support these two patterns also:
** (note that there are no quotes - this would be the default)
drop/dir1/** (this would be given if there is only one spec)
"+drop/dir1/**","-drop/dir1/dir2/**" (no space between first & second)

it is not clear if the below logic takes care of these specs. could you pl. confirm? you can also update the comments to include these specs.

GitHubSriramB (approved):

omeshp:
looks much cleaner!!

omeshp (approved):

PR #314, opened 2017-11-27
log telemetry for older agent

omeshp:


Extensions/ArtifactEngine/Engine/artifactEngine.ts
```diff
@@ -33,11 +33,11 @@ export class ArtifactEngine {
                     this.logger.logSummary();
                     resolve(this.artifactItemStore.getTickets());
                 }, (err) => {
-                    ci.publishEvent('issue', { issueType: 'error', errorMessage: err });
+                    ci.publishEvent('reliability', <ci.IReliabilityData>{ issueType: 'error', errorMessage: err });
                     reject(err);
                 });
             }, (err) => {
-                ci.publishEvent('issue', { issueType: 'error', errorMessage: err });
+                ci.publishEvent('reliability', { issueType: 'error', errorMessage: err });
```
GitHubSriramB: no need to type case to <ci.IReliabilityData> here?

Extensions/ArtifactEngine/Engine/cilogger.ts
```diff
@@ -1,21 +1,45 @@
...
         builduri: tl.getVariable('Build.BuildUri'),
-        buildid: tl.getVariable('Build.Buildid')
+        buildid: tl.getVariable('Build.Buildid'),
+        agentVersion: tl.getVariable('Agent.Version'),
+        version: packagejson.version
     };
 }
 
-export function publishEvent(feature, properties: { [key: string]: any }): void {
+export function publishEvent(feature, properties: any | IReliabilityData): void {
```
GitHubSriramB: why not only IReliabilityData?

Extensions/ArtifactEngine/Engine/cilogger.ts
```diff
@@ -1,21 +1,45 @@
...
+        if (major > 2 || (major == 2 && minor >= 120)) {
+            telemetry = `##vso[telemetry.publish area=${area};feature=${feature}]${JSON.stringify(Object.assign(getDefaultProps(), properties))}`;
+        }
+        else {
+            if (feature === 'reliability') {
+                let reliabilityData = <IReliabilityData>properties;
+                telemetry = "##vso[task.logissue type=error;code=" + reliabilityData.issueType + ";agentVersion=" + tl.getVariable('Agent.Version') + ";taskId=" + area + "-" + packagejson.version + ";]" + reliabilityData.errorMessage
+            }
+        }
+        console.log(telemetry);;
```
GitHubSriramB: i thought you started using task library to log to console. should we use task library here instead?

Extensions/ArtifactEngine/Engine/cilogger.ts
```diff
@@ -1,21 +1,45 @@
...
+    }
+    catch (err) {
     }
```
GitHubSriramB: log a warning to console if any error is caught?

Extensions/ArtifactEngine/Engine/artifactEngine.ts
```diff
@@ -33,11 +33,11 @@ export class ArtifactEngine {
                     this.logger.logSummary();
                     resolve(this.artifactItemStore.getTickets());
                 }, (err) => {
```
GitHubSriramB: is this the only place to catch & log telemetry within artifact-engine?

can you also check in individual tasks to see if they need to log any telemetry?

GitHubSriramB (approved):

Extensions/ArtifactEngine/Engine/artifactEngine.ts
omeshp: not strictly required but better to cast.

Extensions/ArtifactEngine/Engine/artifactEngine.ts
omeshp: yes all the errors should be propagated back here. for task we will need to add telemetry separately.

Extensions/ArtifactEngine/Engine/cilogger.ts
omeshp: the same method is used to publish perf data also, I didn't create an interface for it since it doesn't seem much use currently.

Extensions/ArtifactEngine/Engine/cilogger.ts
omeshp: task lib is used for debug and warning messages to log normal messages we need console.log

PR #325, opened 2017-12-06
use the more recent veson of service bus

paulgmiller:


sureshtadisetty:
@paulgmiller I've merged this project to single VstsServerTaskHelper project and already updated ServiceBus package to 2.0.
So, this change will not be required.

PR #345, opened 2017-12-21
fix download size issue in log

gykuma:


GitHubSriramB (approved):

Extensions/ArtifactEngine/Providers/webProvider.ts
```diff
@@ -52,11 +52,15 @@ export class WebProvider implements models.IArtifactProvider {
                 reject("No downloadUrl available to download the item.");
             }
 
+            var downloadSize: number = 0;
             var itemUrl: string = artifactItem.metadata['downloadUrl'];
             itemUrl = itemUrl.replace(/([^:]\/)\/+/g, "$1");
             this.httpc.get(itemUrl).then((res: httpm.HttpClientResponse) => {
+                res.message.on('data', (chunk) => {
+                    downloadSize += chunk.length;
```
GitHubSriramB: will this handle large file sizes where there would be multiple chunks?

gykuma: Yes. Each file is streamed as chunks, and total file size is determined as cumulative size of chunks.

PR #352, opened 2017-12-28
Improve error message while faling to create directory due to invalid…

omeshp:
… drive or inaccesible fileshare.


Extensions/ArtifactEngine/Providers/filesystemProvider.ts
```diff
@@ -126,17 +128,28 @@ export class FilesystemProvider implements models.IArtifactProvider {
...
+            var dirName: string = path.dirname(folder);
+            if (dirName === folder) {
+                return false;
             }
 
-            this.ensureDirectoryExistence(dirName);
-            fs.mkdirSync(folder);
-            this._createdFolders[folder] = true;
+            if (this.ensureDirectoryExistence(dirName)) {
+                fs.mkdirSync(folder);
```
GitHubSriramB: will this throw in case of error in creating folder?

omeshp: Yes, this will throw. I will add checks to ensure it is cached.

Issue #281, opened 2017-09-28
IISWebAppMgmt: Type Error in Add-SslCert


zivillian:
The IISVersion is read from Registry as number, but converted to string in [Add-SslCert](https://github.com/Microsoft/vsts-rm-extensions/blob/528c04c13dd85314561412329dc83bfb28b0b6a1/Extensions/IISWebAppDeploy/Src/Tasks/IISWebAppMgmt/AppCmdOnTargetMachines.ps1#L186) `[string]$iisVersion`, thus the [comparison](https://github.com/Microsoft/vsts-rm-extensions/blob/528c04c13dd85314561412329dc83bfb28b0b6a1/Extensions/IISWebAppDeploy/Src/Tasks/IISWebAppMgmt/AppCmdOnTargetMachines.ps1#L206) `$iisVersion -ge 8` fails for IIS >= 10.

We found this, while trying to deploy 2 different websites to a shared host with Server 2016 both using ssl on port 443 with SNI enabled. The second deployment fails with 

> SSL Certificate add failed, Error: 183 Cannot create a file when that file already exists.

because the task tries to add a second cert using ipport instead of hostnameport.

616b2f:
Same issue here.

niadak:
Thanks for reporting the issue. We will fix the bug. Please feel free to send a PR with the fix and contribute to this repo if possible.

Toukiletoukan:
Hi, any update on this issue ?
Thanks !

mmajcica:
Same here. Any idea when and if the proposed fix will become part of a release?

MarcoWesterink:
Any progress? Same here after multiple successful deployments it suddenly decided to throw this error.

marcelocelo:
This is not a bug, at least in my version. When configuring the bindings, there is a "Server Name Indication required" option. By checking this option, it will use hostname and not ip when adding the certificate. That solved the problem for me.

basroovers:
I still have this issue with SNI enabled. Why was my change not approved?

Ajay-MS:
@basroovers  I approved the PR. I have few ask can you please resolve them in PR.

Ajay-MS:
@basroovers 
Also, I observed that issue has been fixed in version 2.* of the task already. 

[...]

PR #378, opened 2018-01-25
Fix certificate add #281

basroovers:
Removed data type string for iisVersion


Ajay-MS:
@basroovers 

We apologise for considering this late.

Can you please provide context why string data type has been removed ?

Ajay-MS:
Linking issue with PR. 
https://github.com/microsoft/azure-pipelines-extensions/issues/281

@basroovers 
Can you please increase extension version and task version. Also, resolve the conflicts.

PR #380, opened 2018-01-31
Addressed bugbash comments 

vithati:
- Change cognitive service region from string to picklist.
- Added example in help.
- Add servergate runson property.


Extensions/TwitterSentimentAnalysis/Src/TwitterSentimentAnalysisAzureFunction.txt
```diff
@@ -33,7 +33,13 @@ public static SentimentScore Run(TaskData req, TraceWriter log)
     {
         return new SentimentScore("Invalid Input", -1, -1);
     }
-
+	
+    hashTag = hashTag.Trim();
+    if (hashTag[0] != '#')
+    {
+		hashTag = string.Format("{0}{1}", "#", hashTag, CultureInfo.InvariantCulture);
```
bansalaseem: CultureInfo.InvariantCulture should be just after Format

PR #394, opened 2018-03-07
Fix where error message isn't generated correctly

dneelyep:
A typo in a string causes the task to not print out the environment's computer name correctly. When tested with the current version of the extension, my system displays the text

"<DateTimeHere> ##[error]Microsoft.PowerShell.Commands.WriteErrorException: Deployment on one or more machines failed. System.Exception: Cannot find MsDeploy.exe location. Verify MsDeploy.exe is installed on  and try operation again."

Fix typo which is causing an error message not to display correctly.


Ajay-MS:
@dneelyep 

Thanks for your contribution. 

I have merged your changes with a new PR - https://github.com/microsoft/azure-pipelines-extensions/pull/650 with L0 tests

PR #417, opened 2018-04-13
Fix tests on linux.

omeshp:


Extensions/ArtifactEngine/Engine/artifactEngine.ts
```diff
@@ -85,7 +85,22 @@ export class ArtifactEngine {
...
         if (item.itemType === models.ItemType.File) {
-            if (tl.match([item.path], this.patternList).length > 0) {
+            var pathToMatch = item.path.replace(/\\/g, '/');
+            var matchOptions = {
+                debug: false,
+                nobrace: true,
+                noglobstar: false,
+                dot: true,
+                noext: false,
+                nocase: true,
```
GitHubSriramB: does this mean case sensitive?

Extensions/ArtifactEngine/Engine/artifactEngine.ts
```diff
@@ -85,7 +85,22 @@ export class ArtifactEngine {
...
+                nonull: false,
+                matchBase: false,
+                nocomment: false,
+                nonegate: false,
+                flipNegate: false
+            };
+
+            if (tl.match([pathToMatch], this.patternList, null, matchOptions).length > 0) {
```
GitHubSriramB: don't we need to fix patternList to also not have back-slashes?

Extensions/ArtifactEngine/EngineTests/artifactEngineTests.ts
```diff
@@ -80,7 +96,7 @@ describe('artifactEngine.processItems', () => {
     it('should call getArtifactItem only for artifact items that match include pattern', (done) => {
         var testProvider = new providers.StubProvider();
         var downloadOptions = new engine.ArtifactEngineOptions();
-        downloadOptions.itemPattern = 'path1\\**\npath3\\**\n!path4\\**';
+        downloadOptions.itemPattern = 'path1/**\npath3/**\n!path4/**';
```
GitHubSriramB: keep one test that has back-slash in pattern & make sure it continues to work?

omeshp: already there.

Extensions/ArtifactEngine/Engine/artifactEngine.ts
omeshp: not required.

Extensions/ArtifactEngine/Engine/artifactEngine.ts
omeshp: yes, this should be false (case insensitive), will change

GitHubSriramB (approved):

PR #424, opened 2018-04-26
Pagination in data sources

romil07:
Changing the data source name to a mustache template for fetching the definitions from external tfs


Lovakumar:
Abandoning outdated PR.

PR #440, opened 2018-06-01
Fix taskId for Gates

sureshtadisetty:
Send taskId as jobId for Gates 
emptyGuid as taskId for Gates


madhuv-msft (commented):
looks good.

ServerTaskHelper/DistributedTask.ServerTask.Remote.Common/TaskProgress/TaskClient.cs
```diff
@@ -82,5 +88,19 @@ public void Dispose()
             vssConnection = null;
             taskClient = null;
         }
+
+        private static Guid GetJobId(string hubName, Guid jobId, Guid taskId)
+        {
+            return hubName.Equals("Gates", StringComparison.OrdinalIgnoreCase)
```
madhuv-msft: do you want to make Gates as const and use the same?

PR #441, opened 2018-06-07
Create pagination doc

romil07:


docs/authoring/endpoints/pagination.md
```diff
@@ -0,0 +1,41 @@
...
+     "callbackContextTemplate": "" //set of parameters that need to be used in the next call
+     "callbackRequiredTemplate": "boolean" //template decides if further calls need to be made to service
+     "initialContextTemplate": "" //defines the initial set of query params
+}
+```
+
+Important fields explained:
+- `endpointUrl` : Standard context here. The url that you intend to call.
+- `callbackContextTemplate` : This is the set of parameters that we will need to call this API again. Generally, this will be derived from the response of the previous call. It's supposed to be a set of key-value pairs.
+- `callbackRequiredTemplate` : This is basically a boolean flag which identifies whether further call to API is needed. It does so based on the response of the previous call.
```
omeshp: Add that recalling the API needs to be handled by the client. In our case it will be task editor UI.

docs/authoring/endpoints/pagination.md
```diff
@@ -0,0 +1,41 @@
...
+- `initialContextTemplate` : This is the set of parameters needed for the first call. It is also a set of key-value pairs.
+
+Depending on the kind of pagination supported by the underlying API, the logic to determine the value of "callbackRequiredTemplate" can be anything. 
+We expect both "callbackContextTemplate" and "callbackRequiredTemplate" to be provided as mustache templates. Examples follow:
+
+### Sample Data Source
```
omeshp: Add examples of how 2 well know formats can be supported?

omeshp (approved):

docs/authoring/endpoints/pagination.md
```diff
@@ -0,0 +1,41 @@
+# Pagination
+The support of pagination is data sources is given by following template:
+```
+{
+     "endpointUrl": “someUrl”
+     "callbackContextTemplate": "" //set of parameters that need to be used in the next call
```
GitHubSriramB: remove comments as you are explaining these fields down below

docs/authoring/endpoints/pagination.md
```diff
@@ -0,0 +1,41 @@
+# Pagination
+The support of pagination is data sources is given by following template:
```
GitHubSriramB: typo - … pagination in data source bindings ...

docs/authoring/endpoints/pagination.md
```diff
@@ -0,0 +1,41 @@
...
+- `callbackRequiredTemplate` : This is basically a boolean flag which identifies whether further call to API is needed. It does so based on the response of the previous call.
+- `initialContextTemplate` : This is the set of parameters needed for the first call. It is also a set of key-value pairs.
+
+Depending on the kind of pagination supported by the underlying API, the logic to determine the value of "callbackRequiredTemplate" can be anything. 
+We expect both "callbackContextTemplate" and "callbackRequiredTemplate" to be provided as mustache templates. Examples follow:
+
+### Sample Data Source
+```
+{
+    "endpointUrl": "{{endpoint.url}}/{{project}}/_apis/****/****?&$top=500&continuationToken={{{continuationToken}}}",
```
GitHubSriramB: give a real world example here.

also, have reference to some tasks that you have edited to support pagination from here.

GitHubSriramB (approved):

docs/authoring/endpoints/pagination.md
romil07: I will just add the link of the doc which explains this. Explaining that here would be redundant I think.

GitHubSriramB (approved):

PR #458, opened 2018-07-09
Removing the conditional logic on data source selection for pagination

romil07:


Extensions/ExternalTfs/Src/vss-extension.json
```diff
@@ -313,9 +313,14 @@
             "resultTemplate": "{ Value : \"{{id}}\", DisplayValue : \"{{{name}}}\" }"
           },
           {
+            "endpointId": "tfs:teamfoundation",
```
GitHubSriramB: this should be the input corresponding to endpoint, i.e. connection

PR #460, opened 2018-07-11
use vswhere if possible to detect sqlpackage from vs installation

arjgupta:


TaskModules/powershell/TaskModuleSqlUtility/SqlPackageOnTargetMachines.ps1
```diff
@@ -322,38 +353,68 @@ function LocateSqlPackageInVS([string] $version)
...
-            {
-                $dacVersion = $dacVersionDir.Name
-                $dacFullPath = [System.IO.Path]::Combine($dacVersionDir.FullName, "SqlPackage.exe")
+function Find-VSWhere {
+    $vsWhereLocation = [System.IO.Path]::Combine(${env:ProgramFiles(x86)}, 'Microsoft Visual Studio', 'Installer', 'vswhere.exe')
+    if (Test-Path $vsWhereLocation) {
+        Write-Verbose "vswhere.exe location:'$vsWhereLocation'"
+        return $vsWhereLocation
+    }
+    return ''
```
Ajay-MS: Should we return $null instead

Ajay-MS (approved):

PR #463, opened 2018-07-19
Changes related to the implementation of build definition search on a…

romil07:
…rtifact linking phase of RD creation

### DIFF
--- a/Extensions/ExternalTfs/Src/vss-extension.json
+++ b/Extensions/ExternalTfs/Src/vss-extension.json
@@ -124,6 +124,9 @@
               "isRequired": true,
               "dataType": "string",
               "maxLength": 300
+            },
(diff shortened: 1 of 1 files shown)


Extensions/ExternalTfs/Src/vss-extension.json
```diff
@@ -287,6 +126,23 @@
               "maxLength": 300
             }
           },
+          {
+            "id": "definitionSearch",
```
GitHubSriramB: based on today's discussion, need to update input descriptor for definition to return definitionSearch as property. 

pl. sync up with Satinder to get this done the right way.

romil07: Will take care of it.

PR #464, opened 2018-07-24
Kangaroo endpoint type for OAuth L2 test

divyagandhi01:


Extensions/OAuth2EndpointTest/vss-extension.json
```diff
@@ -0,0 +1,49 @@
+{
+    "manifestVersion": 1.0,
+    "id": "vss-services-kangarool2",
+    "version": "1.0.0",
+    "name": "Kangaroo L2 endpoint type for OAuth2 L2 tests",
+    "publisher": "ms-vscs-rm",
```
GitHubSriramB: can we change publisher? so that we don't publish this to marketplace by mistake.

PR #488, opened 2018-09-07
add default version input in custom artifacts

ajinkya599:
add default version input in custom artifacts (bitbucket, teamcity and external tfs artifacts) with latest and select during release creation type as options.

### DIFF
--- a/Extensions/BitBucket/Src/vss-extension.json
+++ b/Extensions/BitBucket/Src/vss-extension.json
@@ -3,7 +3,7 @@
   "id": "vss-services-bitbucket",
   "name": "Bitbucket artifacts for Release Management",
   "publisher": "ms-vscs-rm",
-  "version": "15.136.1",
+  "version": "15.141.0",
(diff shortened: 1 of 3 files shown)


omeshp (approved):

Extensions/BitBucket/Src/vss-extension.json
```diff
@@ -194,6 +194,33 @@
                   "dataType": "string",
                   "maxLength": 300
               }
+          },
+          {
+            "id": "defaultVersionType",
+            "name": "i18n:Default version",
```
GitHubSriramB: we don't support localization for strings in extensions in this repo. so don't think we need i18n prefix

GitHubSriramB (approved):

Extensions/BitBucket/Src/vss-extension.json
ajinkya599: removed the i18n prefixes

PR #489, opened 2018-09-07
removing bitbucket and external tfs endpoint contributions as they are defined as system contributions

ajinkya599:
removing bitbucket and external tfs endpoint contributions as they are defined as system contributions.


omeshp:
@ajinkya599 What happens if the extension is installed for older versions of TFS where these endpoints are not in-built?

omeshp (approved):

PR #490, opened 2018-09-08
TeamCity versions datasource binding with pagination and larger page size

OlduwanSteve:
This was an optimistic change, which doesn't quite work as I'd hoped - I want all of my TeamCity builds listed, because 

It looks like the 'versions' endpoint ignores the pagination when fetching, and I don't think there's anything I can do about that, but I'd love to be wrong!

The default number of builds returned by TeamCity is 100 so this is still an improvement, and maybe one day the pagination will be supported.

### DIFF
--- a/Extensions/TeamCity/Src/vss-extension.json
+++ b/Extensions/TeamCity/Src/vss-extension.json
@@ -78,11 +78,6 @@
             "resultSelector": "jsonpath:$.buildType[*]"
           },
           {
-            "name": "Builds",
-            "endpointUrl": "{{endpoint.url}}httpAuth/app/rest/buildTypes/{{definition}}/builds/?locator=branch:default:any",
-            "resultSelector": "jsonpath:$.build[?(@.status=='SUCCESS')]"
-          },
-          {
             "name": "LatestBuild",
             "endpointUrl": "{{endpoint.url}}httpAuth/app/rest/buildTypes/{{definition}}/builds?locator=branch:default:any&count=1&status=success",
             "resultSelector": "jsonpath:$.build[*]"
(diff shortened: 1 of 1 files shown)


romil07:
@OlduwanSteve 
Yes, we don't support pagination in version right now. But it should be coming very soon.
Regarding your change I have one question though. I will add a comment in the PR.

Extensions/TeamCity/Src/vss-extension.json
```diff
@@ -208,8 +203,12 @@
           },
           {
             "target": "versions",
-            "dataSourceName": "Builds",
-            "resultTemplate": "{ Value : \"{{id}}\", DisplayValue : \"{{{number}}}\" }"
+            "endpointUrl": "{{endpoint.url}}httpAuth/app/rest/buildTypes/{{definition}}/builds/?locator=branch:default:any,count:2000,start:0{{skip}}",
+            "resultSelector": "jsonpath:$.build[?(@.status=='SUCCESS')]",
+            "resultTemplate": "{ Value : \"{{id}}\", DisplayValue : \"{{{number}}}\" }",
+            "callbackContextTemplate": "{\"skip\": \"{{add skip 2000}}\"}",
+            "callbackRequiredTemplate": "{{ result.nextHref }}",
```
romil07: What does result.nextHref mean here?
Can you share your response structure here for me to validate?

OlduwanSteve: Paginated results in the TeamCity API includes a url for the next and previous pages, but only if there is one:

```
{
  "count": 1000, 
  "href": "/httpAuth/app/rest/buildTypes/MyProject_Default/builds/?locator=branch:default:any,count:1000,start:01",
  "nextHref": "/httpAuth/app/rest/buildTypes/MyProject_Default/builds/?locator=branch:(default:any),count:1000,start:1001", 
  "prevHref": "/httpAuth/app/rest/buildTypes/MyProject_Default/builds/?locator=branch:(default:any),count:1,start:0", 
  "build": [
  ... actual build details go here
  ]
}
```

So the intention was just to check that this property is defined and non-empty.

Lovakumar:
@romil07 abandoning this outdated PR.

PR #500, opened 2018-09-19
Adding retries to the API calls

romil07:


Extensions/ExternalTfs/Src/Tasks/DownloadExternalBuildArtifacts/download.ts
```diff
@@ -149,6 +149,31 @@ async function main(): Promise<void> {
...
+    operation().then((result) => {
+        resolve(result);
+    }).catch((error) => {
+        if (currentRetryCount <= 0) {
+            tl.error(tl.loc("OperationFailed", operationName, error));
+            reject(error);
+        }
+        else {
+            console.log(tl.loc('RetryingOperation', operationName, currentRetryCount));
+            tl.debug(error);
```
romil07: Have added the whole error as the debug log. See if this is ok.

GitHubSriramB: good to log this in console as well as tl.Info as well

GitHubSriramB (approved):

Extensions/ExternalTfs/Src/Tasks/DownloadExternalBuildArtifacts/download.ts
romil07: Info level is not supported by taskLib.

PR #556, opened 2018-12-11
Add 'specific version' type to default version support

omeshp:


Lovakumar (approved):

Extensions/TeamCity/Src/vss-extension.json
```diff
@@ -3,7 +3,7 @@
   "id": "vss-services-teamcity",
   "name": "TeamCity artifacts for Release Management",
   "publisher": "ms-devlabs",
-  "version": "15.142.0",
+  "version": "15.145.0",
```
Lovakumar: Are we following sprint number here?

omeshp: yes

PR #568, opened 2019-01-08
update gulp version

omeshp:


gulpfile.js
```diff
@@ -632,14 +633,13 @@ function compileUIExtensions(extensionRoot) {
     };
 }
 
-gulp.task("build", ["compileNode"], function() {
-});
+gulp.task("build", gulp.series("compileNode"));
 
 gulp.task("handoff", function(cb) {
     handoff(cb);
 });
```
Lovakumar: don't we have call cb() like above?

Lovakumar (approved):

gulpfile.js
omeshp: i only changed the ones where gulp complained.

PR #579, opened 2019-01-24
Adding Guid to the dynamic inventory temp file

scottmishra:
The temp file that is created when using a dynamic inventory has a kind of race condition when running multiple ansible playbook tasks on the same hosted vm concurrently

Adding a GUID to the name should make the file name unique allowing for concurrent jobs to be run

### DIFF
--- a/Extensions/Ansible/Src/Tasks/Ansible/ansibleCommandLineInterface.ts
+++ b/Extensions/Ansible/Src/Tasks/Ansible/ansibleCommandLineInterface.ts
@@ -4,6 +4,7 @@ import * as ansibleUtils from './ansibleUtils';
 import { RemoteCommandOptions } from './ansibleUtils'
 import { ansibleTaskParameters } from './ansibleTaskParameters';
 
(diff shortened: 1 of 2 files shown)


Extensions/Ansible/Src/Tasks/Ansible/ansibleCommandLineInterface.ts
```diff
@@ -4,6 +4,7 @@ import * as ansibleUtils from './ansibleUtils';
 import { RemoteCommandOptions } from './ansibleUtils'
 import { ansibleTaskParameters } from './ansibleTaskParameters';
 
+var guid = require('guid-typescript');
```
sachinma: Can you instead use [uuid](https://www.npmjs.com/package/uuid) instead. That is the one we use in our tasks

scottmishra: @sachinma converted to uuid

PR #582, opened 2019-02-04
Updating the patch version of Ansible task

amaljg:
Updating the patch version of the task (missed in the last change).


Extensions/Ansible/Src/Tasks/Ansible/task.json
```diff
@@ -16,7 +16,7 @@
     "version": {
         "Major": 0,
         "Minor": 0,
-        "Patch": 3
+        "Patch": 4
```
sachinma: Shouldn't you also update the extension version?

PR #592, opened 2019-02-14
Fixing the disabled Ansible tests and enabling them

amaljg:
Fixing the disabled Ansible tests and enabling them.


Extensions/Ansible/Src/Tasks/Ansible/ansibleUtils.ts
```diff
@@ -195,6 +197,10 @@ function beginRequestInternal(request: WebRequest): Promise<WebResponse> {
     });
 }
 
+export function getUniqueInventoryFilePath(): string {
```
sachinma: consider renaming to getTemporaryInventoryFilePath

amaljg: Done.

PR #628, opened 2019-06-11
Enabling Ansible task for Deployment Group as well

vinodkumar3:


Extensions/Ansible/Src/Tasks/Ansible/task.json
```diff
@@ -10,13 +10,14 @@
...
+        "Agent",
+        "DeploymentGroup"
     ],
     "author": "Microsoft Corporation",
     "version": {
         "Major": 0,
         "Minor": 0,
-        "Patch": 4
+        "Patch": 5
     },
```
sachinma: You need to update the extension version as well

vinodkumar3: Done

Extensions/Ansible/Src/Tasks/Ansible/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ansible",
-  "version": "0.0.1",
+  "version": "0.0.2",
```
sachinma: this is not the right location to update the extension version

package-lock.json
```diff
@@ -2242,8 +2242,7 @@
         "ansi-regex": {
```
vinodkumar3: @sachinma package-lock.json is being auto modified. Do you think we should include this because sometimes it may create issues due to wrong installs by npm

sachinma: The changes look simple. We can keep it

sachinma (approved):

PR #653, opened 2019-07-22
Enabled extension for onprem

nishubansal:


bansalaseem:
We should mention that it works 2019.1 onwards only.

bansalaseem (approved):

PR #696, opened 2019-09-24
 Terraform Extension Task Refactoring

AdamCoulterOz:
Aligned the Terraform Extension Tasks to have consistent folder naming and extension task names. Also fixed a few spelling mistakes.

This will break anyone who has setup their YAML pipeline definition with `TerraformTaskV1@0`, and they will need to update to use `Terraform@0`. I think this is acceptable considering it is in preview, definitely would want to fix this while in preview.


AdamCoulterOz:
We've done some

PR #701, opened 2019-10-01
Fixed  readme issue in service now

Anumita:
The change was causing tests to fail.
We can't have https://dev.azure.com hardcoded


vithati (approved):

Extensions/ServiceNow/Src/vss-extension.json
```diff
@@ -3,7 +3,7 @@
   "id": "vss-services-servicenowchangerequestmanagement",
   "name": "ServiceNow Change Management",
   "publisher": "ms-vscs-rm",
-  "version": "4.159.0",
+  "version": "4.159.1",
```
nishubansal: You don't have to bump up the version as this version is not released yet.

nishubansal (changes requested):

PR #711, opened 2019-10-28
Add-Type fails on large environment variables

amccool:
remove Add-Type in IIS psm to pass in definitions using large amount of environment variables


amccool:
I need to get a resolution on this task. Manually editing deployed files from the task just would scale for us.

Also, I lack permissions to see why the CI build failed

Issue #729, opened 2019-11-22
Terraform: bug in TerraformTaskV1 plan task with command options


ahukin:
**Bug**: the terraform plugin TerraformTaskV1 plan task doesn't execute correctly when command options are supplied such as -var or -var-file.

The `terraform plan` command is run twice in this task: once without `-out` and then with `-out`.
The first plan runs correctly but the second is not passed the command options.  The quick fix is to pass the command options to the second plan. e.g.

file: `azure-pipelines-extensions/Extensions/Terraform/Src/Tasks/TerraformTaskV1/src/base-terraform-command-handler.ts` 

```
abstract class BaseTerraformCommandHandler {
...
method: public setOutputVariableToPlanFilePath() {
...
```
**Existing code:**
```
144             let planCommand = new TerraformAuthorizationCommandInitializer( 
145                 "plan", 
146                 tasks.getInput("workingDirectory"), 
147                 tasks.getInput(serviceName, true), 
148                 `-out=${binaryPlanFilePath}` 
149             ); 
```

**Replacement code:**
```
144             let planCommand = new TerraformAuthorizationCommandInitializer( 
145                 "plan", 
146                 tasks.getInput("workingDirectory"), 
147                 tasks.getInput(serviceName, true), 
...                 tasks.getInput("commandOptions"),
148                 `-out=${binaryPlanFilePath}` 
149             ); 
```

I'd fix it but I'm not really a javascript / typescript guy.

Many thanks

Roshan-Kumar-Microsoft:
@ahukin Can you please share debug log to RM_Customer_Queries@microsoft.com

smushtaq:
Facing the same problem. why is it running plan twice. it should be running it only once. Is there any fix on the roadmap?

liamfoneill:
@smushtaq I can't help you with this issue as I am experiencing it myself but if you use '-no-color' in "Additonal Command Arguements" it makes Terraform plan output cleaner as there is some problem in Azure DevOps displaying the colored plus and minus symbols.

smushtaq:
> @smushtaq I can't help you with this issue as I am experiencing it myself but if you use '-no-color' in "Additonal Command Arguements" it makes Terraform plan output cleaner as there is some problem in Azure DevOps displaying the colored plus and minus symbols.

Thanks @liamfoneill for the update. Do you know who can work on this to get it resolved. Should be a quick fix for the team.

omelhus:
[...]

Issue #691, opened 2019-09-14
Terraform: Error after plan if variables required


brent-robinson:
**Extension:** Terraform

When using the **plan** command and passing required variables through **Additional command arguments** the task will succeed, but the plan binary and JSON files are not created.

For example with:
* **Command:** `plan`
* **Additional command arguments:** `-var key_vault_name=mykeyvault123`

The plan will be displayed:

```
##[debug]exec tool: /opt/hostedtoolcache/terraform/0.12.8/x64/terraform
##[debug]arguments:
##[debug]   plan
##[debug]   -var
##[debug]   key_vault_name=votingapp-dev-50735032
[command]/opt/hostedtoolcache/terraform/0.12.8/x64/terraform plan -var key_vault_name=mykeyvault123
Refreshing Terraform state in-memory prior to plan...
The refreshed state will be used to calculate this plan, but will not be
persisted to local or remote state storage.
...
...
```

But the output of the binary plan file will fail:

```
##[debug]exec tool: /opt/hostedtoolcache/terraform/0.12.8/x64/terraform
##[debug]arguments:
##[debug]   plan
##[debug]   -out=/home/vsts/work/r1/a/plan-binary-970e7cc1-563e-47f7-956f-764aa7d6aa48.tfplan

Error: Required variable not assigned

The variable "key_vault_name" is required, so Terraform cannot proceed without a defined value for it.
```

The subsequent show command then fails too:

```
##[debug]exec tool: /opt/hostedtoolcache/terraform/0.12.8/x64/terraform
##[debug]arguments:
##[debug]   show
##[debug]   -json
##[debug]   /home/vsts/work/r1/a/plan-binary-970e7cc1-563e-47f7-956f-764aa7d6aa48.tfplan
Terraform couldn't read the given file as a state or plan file.
The errors while attempting to read the file as each format are shown below.

State read error: Error loading statefile: open /home/vsts/work/r1/a/plan-binary-970e7cc1-563e-47f7-956f-764aa7d6aa48.tfplan: no such file or directory

Plan read error: open /home/vsts/work/r1/a/plan-binary-970e7cc1-563e-47f7-956f-764aa7d6aa48.tfplan: no such file or directory
```

It appears the **additional command arguments** are not passed to the second call of the **plan** command.

solidcloudio:
Seeing this issue too.  Running the command on the build machine works fine, in my case..

[...]

Issue #686, opened 2019-09-06
terraform: make apply consume saved plan


maximbaz:
In our pipeline we execute terraform commands in the following order:

- terraform plan
- terraform apply

In order to make sure that `apply` will do exactly what `plan` has printed, we must save the output of the `plan` and then pass it to the `apply` command.

It seems `plan` (contrary to `onlyPlan`) is actually able to save the result of the `plan` into a file whose path is saved to `jsonPlanFilePath` variable, but it doesn't seem possible to make `apply` consume the plan saved in `jsonPlanFilePath` (?).

Ideally, I'd control it with a new parameter, something like this:

```
- task: ms-devlabs.custom-terraform-tasks.custom-terraform-release-task.TerraformTaskV1@0
  displayName: "Terraform: plan"
  inputs:
    command: plan

- task: ms-devlabs.custom-terraform-tasks.custom-terraform-release-task.TerraformTaskV1@0
  displayName: "Terraform: apply"
  inputs:
    command: onlyApply
    plan: $(jsonPlanFilePath)

```

RoshanKumarMicrosoft:
@maximbaz Are you also not able to do with additional argument ? Can you please ping you skypeId to  RM_Customer_Queries@microsoft.com we will have a quick call to discuss this.

avishnyakov:
@maximbaz , `-out` parameter for plan provides ability to save plan into a file.
> https://www.terraform.io/docs/commands/plan.html#out-path

Terraform tasks have 'Additional Command Line Argument' option which should look like this:
> -no-color -out=$(System.DefaultWorkingDirectory)/my-plan.tf.plan

It will run this one at the end
> terraform plan -out=path

That said, tasks don't have `show` command. This one come handy to render pre-saved terraform plan while passing it as an immutable build artifact between CI/CD. What can we do about this? @RoshanKumarMicrosoft 

Tried to run `terraform show filepath -json` within PS script after task "Terraform init" but weirdly enough, it cannot find initialized early modules failing terraform!

Roshan-Kumar-Microsoft:
Merged PR ..Will release updated version by this week

PR #741, opened 2019-12-05
commandOptions is not passed to setOutputVariableToPlanFilePath()

omelhus:
Hey friends,

I've encountered an issue when using -var-file with TerraFormTaskV1. The `commandOptions` is passed to terraform in `.onlyPlan()`, but not in `setOutputVariableToPlanFilePath()`. This should be solved by prepending `${tasks.getInput("commandOptions")}` to the execution parameters.

Terraform will use the last applied parameter, so by setting the `-out=` parameter last here it will be set even though it's already provided in `commandOptions`.

Keep up the good work!

Related issues: #686, #691, #729.


omelhus:
@RoshanKumarMicrosoft I've emailed you my Skype id if you need to chat 👍

smushtaq:
when should we expect this fix to be released and ready to use?

Roshan-Kumar-Microsoft:
@smushtaq Not released yet but will releases by this week .

PR #775, opened 2020-02-04
Adding the extension contribution for GitHub Packages

romil07:


Extensions/DotnetCoreCLIPackage/Src/readme.md
```diff
@@ -0,0 +1,21 @@
+# Bitbucket&reg; artifacts for Release management
```
kasubram: Bitbucket?

romil07: The read me got copied from Bitbucket. Will remove this.

Issue #704, opened 2019-10-08
IIS Virtual Directory Update not using credentials in file AppCmdOnTargetMachines.ps1


Keith-B:
The file _AppCmdOnTargetMachines.ps1_ contains a function called **Update-VirtualDirectory**. This then calls **Test-Path** without using the authentication credentials passed into the Update-VirtualDirectory function. This causes deployment pipelines in Azure DevOps to fail with 'Access is Denied' when  virtual directories have different logon credentials. The credentials should be passed through to the **Test-Path** cmdlet.

20shivangi:
@Keith-B Can you please share debug logs of the pipeline by setting system.debug to true

20shivangi:
@Keith-B Any update ?

20shivangi:
Feel free to open the issue if face any. Closing this as there is no response.

BlackyTheDruide:
Hi, 

I have the same Problem. The error seems to be on the mentioned file.

https://github.com/microsoft/azure-pipelines-extensions/blob/76d1a318faaa6191d84240f3f28a4954604e0f5b/TaskModules/powershell/TaskModuleIISManageUtility/AppCmdOnTargetMachines.ps1#L618

The Method Test-Path is used without the $physicalPathAuthenticationCredentials.

https://github.com/microsoft/azure-pipelines-extensions/blob/76d1a318faaa6191d84240f3f28a4954604e0f5b/TaskModules/powershell/TaskModuleIISManageUtility/AppCmdOnTargetMachines.ps1#L609

When you configure a virtual Directory with specific credential the Test-Path will fail with an Access Denied.

PR #806, opened 2020-04-16
Add missing -Credential in Test-Path in Update-VirtualDirectory

BlackyTheDruide:
In function "Update-VirtualDirectory" (AppCmdOnTargetMachines.ps1) the given credential are not pass through the Test-Path function. So the result is that when specific credentials for Virtual Directory is given the Task will fail with a "Test-Path : Access is denied" exception. #704


BlackyTheDruide:
Any update on this pull request? Is something missing? I saw that the azure-pipelines-extensions-ci failed, but I can not see why.

PR #818, opened 2020-05-15
Config backend via EnvVars

markjgardner:
Addresses issue microsoft/azure-pipelines-terraform#89 

Makes service connection optional for init.
If no service connection is not specified, credentials are pulled form environment variables instead. This method does not write any secrets to the .terraform/terraform.state file inside the build workspace.


mericstam:
hi, please do the fix in new repo at: https://github.com/microsoft/azure-pipelines-terraform if you still need the functionality

PR #845, opened 2020-08-14
Distributed task webApi bump

tauhid621:
- For DistributedTask.ServerTask.Remote.Common : Bumped Microsoft.TeamFoundation.DistributedTask.WebApi from 15.131.1 to 16.153.0


rohit-batra (approved):
Edited - Oops :) Didn't look into the second part of the fix description.

PR #853, opened 2020-09-03
Added Build visibility, GET call to fetch CR number from context to Update CR Task

pavanvamsi3:
What?
_Same as title_

Why?
Need this task for updating the Change Request in Build Pipeline.

### DIFF
--- a/Extensions/ServiceNow/Src/Tasks/UpdateChangeRequest/UpdateChangeRequestV2/task.json
+++ b/Extensions/ServiceNow/Src/Tasks/UpdateChangeRequest/UpdateChangeRequestV2/task.json
@@ -142,7 +142,7 @@
           },
           "ExecutionOptions": {
             "OutputVariables": "{\"CHANGE_REQUEST_NUMBER\" :  \"jsonpath('$.result[0].number')[0]\", \"CHANGE_CORRELATION_ID\" :  \"jsonpath('$.result[0].correlation_id')[0]\"}",
-            "SkipSectionExpression": "and(eq(isNullOrEmpty(taskInputs['ChangeRequestNumber']), false), eq(eq('$(system.hosttype)', 'build'), true))"
+            "SkipSectionExpression": "eq(isNullOrEmpty(taskInputs['ChangeRequestNumber']), false)"
           }
         },
         {
--- a/Extensions/ServiceNow/Src/Tasks/UpdateChangeRequest/UpdateChangeRequestV2/task.json
+++ b/Extensions/ServiceNow/Src/Tasks/UpdateChangeRequest/UpdateChangeRequestV2/task.json
@@ -16,7 +16,7 @@
   "version": {
     "Major": 2,
     "Minor": 162,
-    "Patch": 1
+    "Patch": 2
   },
   "instanceNameFormat": "Update ServiceNow Change Request",
   "groups": [{
--- a/Extensions/ServiceNow/Src/Tasks/UpdateChangeRequest/UpdateChangeRequestV2/task.json
+++ b/Extensions/ServiceNow/Src/Tasks/UpdateChangeRequest/UpdateChangeRequestV2/task.json
@@ -127,7 +127,8 @@
             "Expression": "eq(count(root['result']), 1)"
           },
           "ExecutionOptions": {
-            "OutputVariables": "{\"CHANGE_CORRELATION_ID\" :  \"jsonpath('$.result[0].correlation_id')[0]\"}"
+            "OutputVariables": "{\"CHANGE_REQUEST_NUMBER\" :  \"jsonpath('$.result[0].number')[0]\", \"CHANGE_CORRELATION_ID\" :  \"jsonpath('$.result[0].correlation_id')[0]\"}",
+            "SkipSectionExpression": "eq(isNullOrEmpty(taskInputs['ChangeRequestNumber']), true)"
           }
         },
         {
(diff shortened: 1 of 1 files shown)


Extensions/ServiceNow/Src/Tasks/UpdateChangeRequest/UpdateChangeRequestV2/task.json
```diff
@@ -126,15 +127,30 @@
             "Expression": "eq(count(root['result']), 1)"
           },
           "ExecutionOptions": {
-            "OutputVariables": "{\"CHANGE_CORRELATION_ID\" :  \"jsonpath('$.result[0].correlation_id')[0]\"}"
```
tauhid621, line 129: Are we going to have ChangeRequestNumber input as required?

pavanvamsi3: Yeah, keeping it as required for release functionality purposes. Users can give or have in empty in YAML.

tauhid621 (approved):

Extensions/ServiceNow/Src/Tasks/UpdateChangeRequest/UpdateChangeRequestV2/task.json
```diff
@@ -126,15 +127,30 @@
...
+            "EndpointId": "$(ServiceNowConnection)",
+            "EndpointUrl": "$(endpoint.url)/api/now/table/change_request?sysparam_query=u_stage_id=$(system.stageId)&u_build_id=$(build.buildId)&u_stage_attempt=$(system.stageAttempt)&sysparm_fields=number,correlation_id",
+            "Method": "GET",
+            "Headers": "{\"Content-Type\":\"application/json\", \"Accept\":\"application/json\"}",
+            "WaitForCompletion": "false",
+            "Expression": "eq(count(root['result']), 1)"
+          },
+          "ExecutionOptions": {
+            "OutputVariables": "{\"CHANGE_REQUEST_NUMBER\" :  \"jsonpath('$.result[0].number')[0]\", \"CHANGE_CORRELATION_ID\" :  \"jsonpath('$.result[0].correlation_id')[0]\"}",
+            "SkipSectionExpression": "and(eq(isNullOrEmpty(taskInputs['ChangeRequestNumber']), false), eq(eq('$(system.hosttype)', 'build'), false))"
```
rohit-batra, line 145: For this to not Skip , ChangeRequest should not be null/empty and HostType should not be Build. I did not get this "Should not be build part"

rohit-batra (changes requested):

rohit-batra (approved):

Extensions/ServiceNow/Src/Tasks/UpdateChangeRequest/UpdateChangeRequestV2/task.json
```diff
@@ -15,7 +16,7 @@
   "version": {
     "Major": 2,
     "Minor": 162,
-    "Patch": 1
```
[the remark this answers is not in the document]
pavanvamsi3: Sounds good, I didn't have an idea about this :v:

Extensions/ServiceNow/Src/Tasks/UpdateChangeRequest/UpdateChangeRequestV2/task.json
pavanvamsi3:
```
Evaluating: and(eq(isNullOrEmpty(taskInputs['ChangeRequestNumber']), False), eq(eq('build', 'build'), False))
				Evaluating and:
				..Evaluating eq:
				....Evaluating isNullOrEmpty:
				......Evaluating indexer:
				........Evaluating taskInputs:
				........=> Object
				........Evaluating String:
				........=> 'ChangeRequestNumber'
				......=> ''
				....=> True
				....Evaluating Boolean:
				....=> False
				..=> False
				=> False
				Expanded: and(eq(True, False), eq(eq('build', 'build'), False))
				Result: False
```

pavanvamsi3: Fixed this.
The evaluation would be (False, True) in case of Build which would not skip. 
This is working as expected because we were fetching the CR number in the other call when the CR number is mentioned. We can skip this call when CR is given and HostType is Build.

pavanvamsi3: Removing the hosttype condition, CR condition will suffice in this case.

If CR number is present (which is always in Release) we'll skip this call. So, this call won't be made in Release at all.

If CR number is present in Build, we'll skip and use the first call, if not we'll make this call and get CR number (in Build).

pavanvamsi3: Verifying this, will retest and merge. Functionality is intact but there's an extra call happening due to this condition.

pavanvamsi3: Removed the host type condition itself, as having a change request is mandatory in Release. We won't need the condition for now.

pavanvamsi3:
Tested the changes post changes 👍 

Good to go.

PR #770, opened 2020-01-20
updated deprecated provider arguments

smspencertx:
This is to resolve warnings from Terraform when using the Azure (ARM) provider backend configuration arguments for subscription_id, tenant_id, client_id, client_secret. There is no need for the arm_ prefix.

https://www.terraform.io/docs/providers/azurerm/index.html

Issue #738, opened 2019-12-02
Update Terraform task 


liamfoneill:
Terraform Task is using deprecated backend config input parameters. Please update them.

Terraform Task Version: 0.0.137
Terraform Version: 0.12.16
Azure Provider Version: 1.37.0

`2019-12-02T16:17:08.0597927Z Initializing the backend...
2019-12-02T16:17:08.1348103Z 
2019-12-02T16:17:08.1349587Z Successfully configured the backend "azurerm"! Terraform will automatically
2019-12-02T16:17:08.1349978Z use this backend unless the backend configuration changes.
2019-12-02T16:17:08.5835303Z 
2019-12-02T16:17:08.5838278Z Warning: "arm_tenant_id": [DEPRECATED] `arm_tenant_id` has been replaced by `tenant_id`
2019-12-02T16:17:08.5838427Z 
2019-12-02T16:17:08.5838521Z 
2019-12-02T16:17:08.5838559Z 
2019-12-02T16:17:08.5838639Z Warning: "arm_subscription_id": [DEPRECATED] `arm_subscription_id` has been replaced by `subscription_id`
2019-12-02T16:17:08.5838753Z 
2019-12-02T16:17:08.5838789Z 
2019-12-02T16:17:08.5838827Z 
2019-12-02T16:17:08.5838946Z Warning: "arm_client_id": [DEPRECATED] `arm_client_id` has been replaced by `client_id`
2019-12-02T16:17:08.5839017Z 
2019-12-02T16:17:08.5839051Z 
2019-12-02T16:17:08.5839086Z 
2019-12-02T16:17:08.5839211Z Warning: "arm_client_secret": [DEPRECATED] `arm_client_secret` has been replaced by `client_secret`
2019-12-02T16:17:08.5839277Z 
2019-12-02T16:17:08.5839362Z 
2019-12-02T16:17:08.5839398Z `

![image](https://user-images.githubusercontent.com/1243172/69977136-2316f580-1522-11ea-95a2-5d9722b2a410.png)

Roshan-Kumar-Microsoft:
@liamfoneill Due to other priority work we are not picking this for now. We are open for contribution. You can create PR and assign to me.

NK74:
Hi,
still not fixed

copjon:
Any movement on approving #863? It would be nice to stop getting deprecation warnings in my builds 😅

AmrutaKawade:
[...]

PR #863, opened 2020-10-03
Remove deprecated arm_ prefix from terraform backend

### LINKED ISSUE #786: Backend Config items deprecated
# Terraform Backend configuration

Following codebase need to update the back configuration. 
`azure-pipelines-extensions/Extensions/Terraform/Src/Tasks/TerraformTaskV1/src/azure-terraform-command-handler.ts`

currently arm_* prefix is used in the backend configuration which are deprecated in terraform resulting in a warning messages. 

```
this.backendConfig.set('arm_subscription_id', tasks.getEndpointDataParameter(backendServiceName, "subscriptionid", true));
this.backendConfig.set('arm_tenant_id', tasks.getEndpointAuthorizationParameter(backendServiceName, "tenantid", true));
this.backendConfig.set('arm_client_id', tasks.getEndpointAuthorizationParameter(backendServiceName, "serviceprincipalid", true));
this.backendConfig.set('arm_client_secret', tasks.getEndpointAuthorizationParameter(backendServiceName, "serviceprincipalkey", true));
```

hfurubotten:
The `arm_` prefix on backend configurations has been deprecated back in 2018 ([see pr for it here](https://github.com/hashicorp/terraform/pull/19448)) and released with terraform `v0.12`. Updated to match latest [documentation](https://www.terraform.io/docs/providers/azurerm/index.html#argument-reference) from terraform which lists current configuration for terraform `v0.12` and `v0.13`.

This is a continuation of PR #770, but with updated tests as well. Hopefully those will go through with success now.

Related issues:
- fixes #786
- #738

### DIFF
--- a/Extensions/Terraform/Src/Tasks/TerraformTaskV1/Tests/InitTests/Azure/AzureInitFailInvalidWorkingDirectory.ts
+++ b/Extensions/Terraform/Src/Tasks/TerraformTaskV1/Tests/InitTests/Azure/AzureInitFailInvalidWorkingDirectory.ts
@@ -30,7 +30,7 @@ let a: ma.TaskLibAnswers = <ma.TaskLibAnswers> {
         "terraform": true
     },
     "exec": {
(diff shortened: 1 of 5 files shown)


AmrutaKawade:
Thanks for the PR. but I think it will break for the users who are using older version of Terraform. I will validate this

timja:
> Thanks for the PR. but I think it will break for the users who are using older version of Terraform. I will validate this

looks like it will break on 0.11.x. see also https://github.com/charleszipp/azure-pipelines-tasks-terraform/issues/139

Is it possible to publish a major version of the task and users who haven't upgraded could pin their version?

sebader:
I agree with @timja. That's what versioning of the tasks is for. Make this change into a new major version and it will not automatically upgrade (and break) for anybody without manually selecting the new version.

hfurubotten:
Closing this as #889 supersedes this PR and provides new version.

PR #871, opened 2020-10-20
Added image references

pavanvamsi3:
What?
Added image references for marketplace description.
https://marketplace.visualstudio.com/items?itemName=ms-vscs-rm.vss-services-servicenowchangerequestmanagement

Why?

For the images to be included in the `vsix` file, we need to add a reference for the image path  in the `vss-extension.json.`

Also, updated the version number 'cos `4.176.0` is already in the marketplace.

Image from VSIX Viewer:

files


tauhid621 (approved):

tauhid621:
We can install extension in tfat and check there if these images appear there or not.

PR #875, opened 2020-10-22
Fix terraform plan so it runs only once

ynden:
#### The problem
Actually, the terraform plan command runs twice. 
The first time it runs with no output, and if terraform version is >= 0.12.0, it will run a second time -- but this time, with an output.

#### Solution
I modified the plan function, so it runs **terraform plan** with *output* if terraform is >= *0.12.0*, else it will only run plan.

#### Additional informations
Fixes #848


ynden:
@romil07 @tauhid621 can you please review, thank you!

ynden (commented):
I modified the plan function, so it runs **terraform plan** with *output* only if terraform is >= *0.12.0*, else it will only run plan.
With this fix, we can gain some time, instead of re-running terraform plan.

mericstam:
fixed in V3 in  new repo at: https://github.com/microsoft/azure-pipelines-terraform

Issue #851, opened 2020-08-27
IIS Web app Manage: cannot detect existing apppool, site and binding if they contains foreign characters


AThomsen:
Having a website with matching appool and bindings with names that contains foreign characters (in my case a danish "æ"), causes subsequent deploys to fail as the existing apppool, website and bindings are not detected and so the task tries to create these again.

It works for me by adding ` -Encoding ([System.Text.Encoding]::GetEncoding(1252))` to all `Invoke-VstsTool` calls in `AppCmdOnTargetMachines.ps1`. 
Not sure if that solution is safe world-wide. I guess that depends on how appcmd.exe is constructed.

AmrutaKawade:
@AThomsen Feel free to raise PR if you know the solution. We highly encourage community contribution.

20shivangi:
@AThomsen Thanks for your contribution. We will review this

PR #884, opened 2020-11-13
invoke appcmd with encoding 1252

AThomsen:
Having a website with matching appool and bindings with names that contains foreign characters (in my case a danish "æ"), causes subsequent deploys to fail as the existing apppool, website and bindings are not detected and so the task tries to create these again.

It works for me by adding -Encoding ([System.Text.Encoding]::GetEncoding(1252)) to all Invoke-VstsTool calls in AppCmdOnTargetMachines.ps1.
Not sure if that solution is safe world-wide. I guess that depends on how appcmd.exe is constructed.

Bug filed as #851

### DIFF
--- a/TaskModules/powershell/TaskModuleIISManageUtility/AppCmdOnTargetMachines.ps1
+++ b/TaskModules/powershell/TaskModuleIISManageUtility/AppCmdOnTargetMachines.ps1
@@ -43,7 +43,7 @@ function Test-WebsiteExist
     $command = "`"$appCmdPath`" $appCmdArgs"
     Write-Verbose "Checking website exists. Running command : $command"
 
-    $website = Invoke-VstsTool -Filename $appCmdPath -Arguments $appCmdArgs
+    $website = Invoke-VstsTool -Filename $appCmdPath -Arguments $appCmdArgs -Encoding ([System.Text.Encoding]::GetEncoding(1252))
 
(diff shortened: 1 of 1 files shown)


TaskModules/powershell/TaskModuleIISManageUtility/AppCmdOnTargetMachines.ps1
```diff
@@ -43,7 +43,7 @@ function Test-WebsiteExist
     $command = "`"$appCmdPath`" $appCmdArgs"
     Write-Verbose "Checking website exists. Running command : $command"
 
-    $website = Invoke-VstsTool -Filename $appCmdPath -Arguments $appCmdArgs
+    $website = Invoke-VstsTool -Filename $appCmdPath -Arguments $appCmdArgs -Encoding ([System.Text.Encoding]::GetEncoding(1252))
```
arjgupta, line 46: @AThomsen instead of making this ::GetEncoding(1252), can this instead be [System.Text.Encoding::DefaultEncoding

AThomsen: It works for me by adding 1252. Haven't tried anything else.

Microsoft is really the one to ask about this as the correct solution requires insight in how "appcmd" (closed source, MS) is put together.

PR #905, opened 2021-01-22
[ArtifactEngine] Add preview version for Node 10 migration

EzzhevNikita:
Prepared ArtifactEngine preview version for publishing to NPM. This is required for Node 10 migration of DownloadBuildArtifactsV0 task

What was changed:
* Updated dependencies; Removed unused definitions 
* Changed typescript version to 4.0.2;
* Changed deprecated assert.notEqual and assert.equal to assert.notStrictEqual and assert.strictEqual;
* Bumped package version;
* Removed deprecated @types/handlebars reference, since handlebars lib has already build-in type definitions

**The library was tested as a part of DownloadBuldArtifactsV0 task with the next cases** ([pipeline reference](https://dev.azure.com/v-niezz-testorg/TestProject/_build/results?buildId=2070&view=results)):
* Download FileShare artifact on windows-latest agent
* Download Container artifact on windows-latest agent
* Download Container artifact on ubuntu-latest agent
* Download Container artifact on macOS-latest agent 

**NOTE**: It needs to change the target branch since this is a preview version of the package. But we don't have enough permissions for it.


Extensions/ArtifactEngine/package.json
```diff
@@ -18,23 +18,23 @@
   "dependencies": {
     "handlebars": "4.0.14",
     "minimatch": "3.0.2",
-    "azure-pipelines-task-lib": "2.8.0",
+    "azure-pipelines-task-lib": "^3.0.6-preview.1",
     "tunnel": "0.0.4"
   },
   "devDependencies": {
-    "@types/handlebars": "^3.0.28",
+    "@types/handlebars": "^3.0.3",
```
DaniilShmelev, line 25: Why do we need to change this?

anatolybolshakov (approved):
LGTM, but please take a look at Daniil's comment, and let's check all places in in-the-box tasks repo - where this package is imported, and how version is specified

Extensions/ArtifactEngine/package.json
EzzhevNikita: Truly speaking bumped it by mistake, since it will be better to remove this types reference. As I could see it is marked in [npm](https://www.npmjs.com/package/@types/handlebars) as deprecated because [handlebars](https://www.npmjs.com/package/handlebars) is already contains in-build type definitions. 
So I will remove this reference.

DaniilShmelev (approved):

harshiniks (approved):

narula0781:
Thanks @pavanvamsi3 and @chandra

PR #906, opened 2021-01-22
Fix issue 903

### LINKED ISSUE #903: ServiceNow extensions query service now api wrong
Hi 

When using UpdateChangeRequestV2 to close change in service now.
It fail if the build have multi stages.

Lets say we have stage: "TEST" and "PROD"
Then it work fin for creating and closing change for TEST but when it try to close PROD it fail because the API call to service now return change number for TEST and PROD

The error happen here:
https://github.com/microsoft/azure-pipelines-extensions/blob/master/Extensions/ServiceNow/Src/Tasks/UpdateChangeRequest/UpdateChangeRequestV2/task.json
Line: 137

It can be fixed by change `sysparam_query` to `sysparm_query` and the correct encode the `x_mioms_azpipeline_stage_id, x_mioms_azpipeline_build_id and x_mioms_azpipeline_stage_attempt`

Assume the line should look like this:
`            "EndpointUrl": "$(endpoint.url)/api/now/table/change_request?sysparm_query=x_mioms_azpipeline_stage_id=$(system.stageId)^x_mioms_azpipeline_build_id=$(build.buildId)^x_mioms_azpipeline_stage_attempt=$(system.stageAttempt)&sysparm_fields=number,correlation_id",`

I assume the reason for it is working with just 1 stage is because
service now ignore `sysparam_query` query parameter and only filter for query parameters `x_mioms_azpipeline_build_id` and `x_mioms_azpipeline_stage_attempt`

yellowpanda:
Resolves #903.

### DIFF
--- a/Extensions/ServiceNow/Src/Tasks/UpdateChangeRequest/UpdateChangeRequestV2/task.json
+++ b/Extensions/ServiceNow/Src/Tasks/UpdateChangeRequest/UpdateChangeRequestV2/task.json
@@ -134,7 +134,7 @@
         {
           "RequestInputs": {
             "EndpointId": "$(ServiceNowConnection)",
-            "EndpointUrl": "$(endpoint.url)/api/now/table/change_request?sysparam_query=x_mioms_azpipeline_stage_id=$(system.stageId)&x_mioms_azpipeline_build_id=$(build.buildId)&x_mioms_azpipeline_stage_attempt=$(system.stageAttempt)&sysparm_fields=number,correlation_id",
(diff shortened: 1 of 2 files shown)


pavanvamsi3:
@yellowpanda Thanks, the PR looks good. This will need some testing, I'll update you on that.

pavanvamsi3 (approved):
LGTM 👍 

@yellowpanda Can you please update the task minor version to `183` in this file?

Also update the version as `4.183.1` [here](https://github.com/microsoft/azure-pipelines-extensions/blob/master/Extensions/ServiceNow/Src/vss-extension.json#L6).

pavanvamsi3:
- [ ] Update the version - @pavanvamsi3

pavanvamsi3 (changes requested):
- [x] Update the extension version

yellowpanda:
@pavanvamsi3 Done

pavanvamsi3 (approved):
🌟

PR #889, opened 2020-11-26
Terraform V2

aristosvo:
> Credits to @hfurubotten for the description

The arm_ prefix on backend configurations has been deprecated back in 2018 and released with terraform v0.12. Updated to match latest documentation from terraform which lists current configuration for terraform v0.12 and v0.13 (v0.14 is around the corner

This is a continuation of PR #770 and #863, but with the updates in a new taks version. Hopefully those will go through with success now.

Related issues:
fixes #882

PR #920, opened 2021-03-11
Update azure-terraform-command-handler.ts

terryjrjoyce:
The use of arm_client_id, arm_client_secret, arm_subscription_id and arm_tenant_id will soon be deprecated from the Terraform Provider in favour of those values without the arm_ in the value.  

This change will remove the warning when running the plugin and also provide future support.


aristosvo:
Simular to #889, with the difference that #889 is a new version to prevent breaking changes for users of older versions

DenWin:
Just saying, but these fields will not be deprecated anymore with Terraform 0.15 - they will be GONE!

terryjrjoyce:
This is now urgent as v15 has now been released and this plugin does not work with it.  Means I can't update to the latest version.

mjiderhamn:
See https://github.com/microsoft/azure-pipelines-extensions/pull/958

mericstam:
fixed in new repo at: https://github.com/microsoft/azure-pipelines-terraform

PR #922, opened 2021-03-19
[ArtifactsEngine] Updated handlebars to fix vulnerabilities

egor-bryzgalov:
Updated the handlebars package for the ArtifactsEngine package to resolve vulnerabilities for dependent pipeline tasks.


anatolybolshakov (approved):
LGTM, but we need to test these changes also with tasks which use artifact engine

PR #928, opened 2021-04-01
Fixing extension from .png to .PNG

bauca:
This commits fixes broken images on the readme.md file.
In the repository the images were created with .PNG (uppercased) prefix, changing them to .png (lowercase) might brake other repositories that are relying on them, therefore the best possible solution is to change the reference in the Readme.md file itself.


bauca (commented):
Ideally would be better changing the suffix of the .PNG files themselves to .png, but it might brake references to this repository.

PR #938, opened 2021-04-09
Fix documentation

joaomoreno:
`terraform apply tfplan -auto-approve` fails with:

```
Too many command line arguments. Configuration path expected.
```

The right syntax is `terraform apply -auto-approve tfplan`


mericstam:
This should fixed in V3 in new repo at: https://github.com/microsoft/azure-pipelines-terraform

PR #988, opened 2021-10-21
Terraform task: propogate no-color to validate command when using validate+apply

vexingpos:
When using the apply command in the TF task, it actually performs a validate+apply

However, when you pass in `-no-color` as an additional argument, this is only consumed by the apply command and not by the validate command.

This change will pass in `-no-color`  to the validate command when it is specified for the apply command


mericstam:
Looks good. I will add this PR to our release schedule

PR #1061, opened 2022-05-24
Bumped mocha version in Extensions/ArtifactEngine

sbalia:
Bumped mocha version to 5.2.0 as it will eliminate the vulnerability in the growl package.
5.2.0 is compatible with gulp-mocha, which is defined in the root package.json file.

Affected versions of growl do not properly sanitize input prior to passing it into a shell command, allowing for arbitrary command execution.


sbalia:
Duplication of https://github.com/microsoft/azure-pipelines-extensions/pull/1044

PR #1122, opened 2023-02-03
2023646 DTS: Re-release extension manifest, up versions

z00sts:
DTS: https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/2023646
Extension: https://marketplace.visualstudio.com/items?itemName=ms-vscs-rm.vss-services-bitbucket

I've found that extension manifest still has "old" value for the `endpointTypeId` field (which must be equal to "bitbucket" now, but it equals to "bitbucket:UsernamePassword"):
https://ms-vscs-rm.gallery-assets.vsts.me/_apis/public/gallery/publisher/ms-vscs-rm/extension/vss-services-bitbucket/15.157.0/assetbyname/Microsoft.VisualStudio.Services.Manifest 

We request manifest data this way from backend and expect it to be up-to-date.

So the idea of this PR is to bump extension versions and re-release it to update manifest values here [https://ms-vscs-rm.gallery-assets.vsts.me/_apis/public/gallery/publisher/ms-vscs-rm/extension/vss-services-bitbucket/15.157.0](https://ms-vscs-rm.gallery-assets.vsts.me/_apis/public/gallery/publisher/ms-vscs-rm/extension/vss-services-bitbucket/15.157.0/assetbyname/Microsoft.VisualStudio.Services.Manifest%C2%A0). **There is no need to change something in extension manifest.**

### DIFF
--- a/Extensions/BitBucket/Src/vss-extension.json
+++ b/Extensions/BitBucket/Src/vss-extension.json
@@ -3,7 +3,7 @@
   "id": "vss-services-bitbucket",
   "name": "Bitbucket artifacts for Release Management",
   "publisher": "ms-vscs-rm",
-  "version": "15.157.0",
+  "version": "15.157.1",
   "public": true,
   "description": "Tools related to connecting with Bitbucket",
   "_description.comment": "The below format to define artifact extensions is currently in preview and may change in future.",


Extensions/BitBucket/Src/vss-extension.json
```diff
@@ -1,9 +1,9 @@
 {
-  "manifestVersion": 1.0,
+  "manifestVersion": 2.0,
   "id": "vss-services-bitbucket",
   "name": "Bitbucket artifacts for Release Management",
   "publisher": "ms-vscs-rm",
-  "version": "15.157.0",
+  "version": "15.157.1",
```
z00sts, line 6: Not sure that we have to release "new version" of an extension. Probably just manifestVersion will be enough.

z00sts: According to Scott:

```
the manifestVersion value is not the version of the extension - it defines the format of the file itself.  so we don't want to change that.  what we want is to change the version , which is the version of the extension and what the gallery/EMS will look at to determine whether it changed.  in your PR you have it bumped to 15.157.1, which should be sufficient.
so i think you should undo the change to manifestVersion and then we can merge your PR.  after that we'll have to update it in the gallery
```

So I've rolled back manifestVersion change.

bsmid (approved):

Issue #1058, opened 2022-05-20
Windows Server 2022 is totally broken trying to add a binding


jiggybyte:
The hard coded lines used in the $result.Get(X) are broken in Windows Server 2022, and have been for several months now.  This is a pretty big deal, considering the push in Azure to use the latest images.  No attention to this issue from the authors, it's a five minute fix.

jabteles:
+1 preventing to migrate to Server 2022. No more workarounds.

v-schhabra:
Hi @jabteles,
The mentioned issue has been fixed in this PR https://github.com/microsoft/azure-pipelines-extensions/pull/1220
Could you please use the IISWebAppManagementV3 version of the task and test if it is working fine?

jabteles:
> Hi @jabteles, The mentioned issue has been fixed in this PR #1220 Could you please use the IISWebAppManagementV3 version of the task and test if it is working fine?

I updated the extension on my Azure DevOps and still get the same error. I've confirmed it is using correct version (3.2.0)

v-schhabra:
@jabteles Could you please share the debugged logs?

Issue #1008, opened 2021-11-26
IISWebAppManagementOnMachineGroupV0 can't detect existing SSL bindings on Server 2022


altwohill:
I have the following in my yaml pipeline:

```yaml
  - task: IISWebAppManagementOnMachineGroup@0
    displayName: 'Deploy base website with binding: ${{ binding }}'
    inputs:
      IISDeploymentType: 'IISWebsite'
      ActionIISWebsite: 'CreateOrUpdateWebsite'
#... snipped 
      AddBinding: true
      Protocol: 'https'
      IPAddress: '*'
      Port: 443
      ServerNameIndication: true
      HostNameWithSNI: '${{ binding }}'
```
It works the first time but fails on repeat deployments on Server 2022 machines.

The issue doesn't occur on our Server 2019 and older machines.

I think the issue is due to some additional lines in the netsh output
<img width="1065" alt="image" src="https://user-images.githubusercontent.com/629254/143522186-2dc2e984-6eb7-466e-bad5-f24b94238413.png">

Looking at https://github.com/microsoft/azure-pipelines-extensions/blob/78bf251236c84c4da622c125764b7e7434e98a2a/TaskModules/powershell/TaskModuleIISManageUtility/AppCmdOnTargetMachines.ps1#L186 and https://github.com/microsoft/azure-pipelines-extensions/blob/78bf251236c84c4da622c125764b7e7434e98a2a/TaskModules/powershell/TaskModuleIISManageUtility/AppCmdOnTargetMachines.ps1#L201 I see the script is checking specific line numbers before proceeding. 

On Server 2022 the checks always fail so it tries to add a new binding, which fails with the error

```
SSL Certificate add failed, Error: 183
Cannot create a file when that file already exists.


##[error]Process 'netsh' exited with code '1'.
```

altwohill:
Ah, I see #993 would fix this

danspam:
Just spent hours trying to figure out why the IIS web app manage task was failing. Totally broken on Server 2022

jiggybyte:
This is still completely broken on Server 2022.  Such a shame, there's a simple fix to the script (or just don't use hard coded lines, sheesh)

saephraim:
I just ran into this same issue today.  We have installed Windows Server 2022 to migrate sites to.  This worked fine in Server 2012 R2.

jabteles:
+1 preventing to migrate to Server 2022. No more workarounds.

[...]

PR #1134, opened 2023-03-13
Updated appcmdontargetmachines.ps1

jc-odot:
This resolves an issue with Add-SslCert and Windows 2022. the `netsh` command on 2022 has a different number of lines being output, which results in checks for existing certificate or bindings to fail.

This should resolve following issues: 
#1058 
#1008


mmrazik:
Just one note here -- we most likely can't break backwards compatibility on older windows releases.

joeywas:
> Just one note here -- we most likely can't break backwards compatibility on older windows releases
How far back do you need to confirm backwards cmpatibility?

Issue #919, opened 2021-03-10
SSL Certificate add failed, Error: 183 Cannot create a file when that file already exists


jabteles:
Hi,

Having the same issue as reported in https://github.com/microsoft/azure-pipelines-extensions/issues/623, we created a new certificate, same subject name, different thumbprint.

In the task, only changed the thumbprint for the new cert and the task fails with the error "SSL Certificate add failed, Error: 183 Cannot create a file when that file already exists"

I attach the Task [verbose_log](https://github.com/microsoft/azure-pipelines-extensions/files/6117665/verbose_log.txt)

dirtygooback:
I am also having this same problem.

jabteles:
Would be great to have news on this

iambigred:
We are encountering the same issue.  Seems like such a fundamental problem that has been unresolved for years.

It's been raised with MS since 2019: https://developercommunity.visualstudio.com/t/iis-web-app-manage-task-should-replace-the-certifi/547395

lzocateli:
I'm having the same problem, does anyone have a workaround?

jabteles:
> I'm having the same problem, does anyone have a workaround?

Removing the certificate (PowerShell) before executing this task

yepeekai:
I found the problem.  The output of "netsh http show sslcert" is different in new version of windows server vs older ones.  The number of empty lines at the beginning varies.

[Problematic file](https://github.com/microsoft/azure-pipelines-extensions/blob/master/TaskModules/powershell/TaskModuleIISManageUtility/AppCmdOnTargetMachines.ps1)

The following line is there twice in the script and assume that the relevant information is at line 4 but for newer windows, the information is at line 6.  I think this should be modified and be more robust.
`$isItSameBinding = $result.Get(4).Contains([string]::Format("{0}:{1}", $hostname, $port))`

I don't have time in the short term but the idea would be either to skip empty lines or having a regex searching for the relevant line.

coolhome:
Created this report bug - please upvote!

https://developercommunity.visualstudio.com/t/IISWebAppMgmt3---SSL-Certificate-add-fa/10371316

DrMueller:
Hello there
Got here because I'm currently wrestling with this error. Does anyone have an idea how to circumvent it for the time being?

krilbe:
[...]

PR #1220, opened 2024-01-12
IISWebAppMgmtV3 - Bypass bindings with duplicate certificates & support multiple certificates

v-venunayira:
**Description**: Bypassing addition of SSL binding with an existing certificate.

**Documentation changes required:** (Y/N) <Please mark if documentation changes are required>

**Added unit tests:** (Y/N) <Please mark if unit tests were added or updated according changes>

**Attached related issue:**  https://github.com/microsoft/azure-pipelines-extensions/issues/919

**Checklist**:
- [ ] Gulp build
- [ ] Gulp test

### DIFF
--- a/Extensions/IISWebAppDeploy/Src/Tasks/IISWebAppMgmt/IISWebAppMgmtV1/task.json
+++ b/Extensions/IISWebAppDeploy/Src/Tasks/IISWebAppMgmt/IISWebAppMgmtV1/task.json
@@ -16,7 +16,7 @@
   "version": {
     "Major": 1,
     "Minor": 4,
-    "Patch": 6
+    "Patch": 7
   },
   "demands": [
   ],
--- a/Extensions/IISWebAppDeploy/Src/Tasks/IISWebAppMgmt/IISWebAppMgmtV2/task.json
+++ b/Extensions/IISWebAppDeploy/Src/Tasks/IISWebAppMgmt/IISWebAppMgmtV2/task.json
@@ -16,7 +16,7 @@
   "version": {
     "Major": 2,
     "Minor": 2,
-    "Patch": 6
+    "Patch": 7
   },
   "demands": [
   ],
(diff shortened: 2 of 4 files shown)


v-mohithgc (approved):
LGTM, but @LeftTwixWand can you confirm if bumping of task version for V1 and V2 is required?

LeftTwixWand (approved):

LeftTwixWand:
> LGTM, but @LeftTwixWand can you confirm if bumping of task version for V1 and V2 is required?
Hey @v-mohithgc it's not required, but it would be good to have the same version for all the tasks, just not to mess up in case of a problem.

Extensions/IISWebAppDeploy/Src/vss-extension.json
```diff
@@ -2,7 +2,7 @@
   "manifestVersion": 1,
   "extensionId": "iiswebapp",
   "name": "IIS Web App Deployment Using WinRM",
-  "version": "1.6.8",
+  "version": "1.235.0",
```
LeftTwixWand, line 5: Actually, I don't see much need to bump the version to the current sprint.
It might be useful in the tasks repository because there is active work on each task and because this approach has been used for a long time.
But here it looks inconsistent and confusing for customers: Why was the version bumped from 6 to 235.

So, I'd suggest just going with the normal increase of the minor versions.

v-venunayira: Done

LeftTwixWand (approved):

jabteles:
Hey @LeftTwixWand and thanks for this fix! But when will it be available as a release in the Azure/Visual Studio Marketplace so we can download to our on-premises installations?

LeftTwixWand:
Hello @jabteles we've just released a new version of the extension.
The ADO should automatically catch up the update of the extension and install it.

jabteles:
> Hello @jabteles we've just released a new version of the extension. The ADO should automatically catch up the update of the extension and install it.

I updated the extension on my Azure DevOps and still get the same error

naillatem:
The extension fails on target server from other localizations. Parsing the **nesth http show sslcert** command output with an unlocalized string, like **"IP:port"** doesn't work on target servers with different localization configuration (i.e. spanish output shows **"IP:puerto"**, and **Add-SslCert** function fails when trying to add an existing certificate binding).

PhilBlackRose:
@v-schhabra Thanks for hint to try this to fix issues with IIS web deploy task, This task however doesn't seem compatible with Deployment Group jobs?

v-schhabra:
> > Hello @jabteles we've just released a new version of the extension. The ADO should automatically catch up the update of the extension and install it.
> 
> I updated the extension on my Azure DevOps and still get the same error. I've confirmed it is using correct version (3.2.0)

@jabteles Could you share the logs of the pipeline by adding variable system.debug to "true"

PR #1276, opened 2025-08-05
Add gulp build --test

LeftTwixWand:
**Description**: <Describe your changes here>

Added --test argument to gulp build command.
This argument marks, that the build should produce test case extensions for e2e testing and private releases. 
After the `gulp build --test` and `gulp package` all the extensions should have updated id and visibility set to private:
image

**Added unit tests:** N
**Attached related issue:** https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/2285466

**Checklist**:

### DIFF
--- a/.pipelines/1es-migration/azure-pipelines.yml
+++ b/.pipelines/1es-migration/azure-pipelines.yml
@@ -64,10 +64,14 @@ extends:
     sdl:
       sourceAnalysisPool:
         name: 1ESPtTfsAgentBuildPoolSDL
-      spotBugs: { enabled: false }
-      credscan: { enabled: true }
-      binskim: { enabled: true }
-      eslint:  { enabled: true }
+      spotBugs:
+        enabled: false
+      credscan:
+        enabled: true
+      binskim:
+        enabled: true
+      eslint:
+        enabled: true
 
     pool:
       name: 1ESPtTfsAgentBuildPool1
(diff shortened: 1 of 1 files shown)


starkmsu (approved):

.pipelines/1es-migration/azure-pipelines.yml
```diff
@@ -92,151 +80,242 @@ extends:
...
+          inputs: { command: 'custom', customCommand: 'install -g gulp-cli' }
 
         - task: PowerShell@2
-          displayName: 'Build Extensions with Gulp'
-          inputs:
-            targetType: 'inline'
-            script: 'gulp build'
-            workingDirectory: '$(Build.SourcesDirectory)'
+          displayName: 'Build MAIN - gulp build'
+          inputs: { targetType: inline, script: 'gulp build', workingDirectory: '$(Build.SourcesDirectory)' }
```
nemanjarogic, line 123: Any reason why you switched to this syntax? Yaml files usually follow the approach below and it's more readable or easier to extend if we need to add new properties. I commented this line, but it applies to the whole file.

```
- task: PowerShell@2
          displayName: 'Build Extensions with Gulp'
          inputs:
            targetType: 'inline'
            script: 'gulp build'
            workingDirectory: '$(Build.SourcesDirectory)'
```

LeftTwixWand: You're right, that standard yaml is more readable. Fixed it👌

rm-dmiri (approved):

PR #1365, opened 2026-04-23
Remove Q dependency and refactor promise handling in task validation

ivanduplenskikh:
**Description**: This PR removes `q` dependency from the root project

**Added unit tests:** N

**Attached related issue:** N

**Checklist**:


package.js
```diff
@@ -76,114 +82,111 @@ function parseTaskJson(jsonContents) {
...
+            // Statically link the required internal common modules.
+            let taskDeps;
+
+            if ((taskDeps = commonDeps[folderName])) {
+                taskDeps.forEach(function (dep) {
+                    gutil.log('Linking ' + dep.module + ' into ' + folderName);
+                    const src = path.join(commonSrc, dep.module, "Src/");
+                    const dest = path.join(targetPath, dep.dest);
+                    shell.mkdir('-p', dest);
+                    fs.copy(src, dest, function (err) {
```
manolerazvan, line 125: Old code passed **"*"** as a filter argument: fs.copy(src, dest, "*", cb). That 3rd arg was silently removed. If it was acting as a glob filter in the fs-extra version used here, this now copies all files instead of only matching ones. 

Was this change intentional ?

package.js
```diff
@@ -76,114 +82,111 @@ function parseTaskJson(jsonContents) {
  * @returns {NodeJS.ReadWriteStream} A through2 stream that processes task.json files, copies common modules, and builds UI contributions.
  */
 function copyCommonModules(currentExtnRoot, commonDeps, commonSrc, extensionSourcePath) {
-    return through.obj(
-        function(taskJson, _encoding, done) {
-            if (!fs.existsSync(taskJson)) {
-                new gutil.PluginError('PackageTask', 'Task json cannot be found: ' + taskJson.path);
+    return through.obj(function(taskJson, _encoding, done) {
+        if (!fs.existsSync(taskJson)) {
+            new gutil.PluginError('PackageTask', 'Task json cannot be found: ' + taskJson.path);
```
manolerazvan: the error is created but never thrown or passed to done(). This bug existed before this PR too, but since the surrounding code was touched, it's a good opportunity to fix it to return done(createError(...)).

package.js
ivanduplenskikh: It was intentional.
The reason why I removed the third argument is that [`@types/fs-extra@8.0.0`](https://www.npmjs.com/package/@types/fs-extra/v/8.0.0?activeTab=code) doesn't contain a definition with argument where the third is a glob filter

```
export function copy(src: string, dest: string, options?: CopyOptions): Promise<void>;
export function copy(src: string, dest: string, callback: (err: Error) => void): void;
export function copy(src: string, dest: string, options: CopyOptions, callback: (err: Error) => void): void;
```

I've checked `fs-extra@8.1.5` that is used in package-lock.json and it doesn't handle the third string argument [here](https://github.com/jprichardson/node-fs-extra/blob/b7df7cce3f7ca5bc0ab85110aa997bd0ad33482f/lib/copy/copy.js#L11-L15)

package.js
ivanduplenskikh: Fixed, thank you

manolerazvan (approved):

rm-dmiri (approved):

PR #1367, opened 2026-04-24
Enable unit tests in CI

ivanduplenskikh:
**Description**: This PR enable unit tests in CI

- Improved error handling in WebProvider to ensure proper rejection on invalid JSON response which was a cause of fail in unit tests.
- Updated unit tests for filesystemProvider and webProvider to handle asynchronous operations and mock file system interactions more effectively.
- Modified localization strings across multiple languages to reflect the new "artifact-engine-v2" naming convention.
- Enhanced CI pipeline configuration to allow running all test suites and improved the logic for determining which tests to execute based on changed files.
- Removed outdated `getUpdatedPaths.js` script and refactored gulp tasks for better clarity and functionality.

**Added unit tests:** N

**Attached related issue:** N

**Checklist**:


Extensions/Ansible/Tests/Tasks/Ansible/mockAnsibleUtils.ts
```diff
@@ -33,10 +33,10 @@ export function runCommandOnRemoteMachine(cmd: string, sshClient: any, options:
     return defer.promise;
 }
 
-export function copyFileToRemoteMachine(scriptFile: string, scpConfig: any): Q.Promise<string> {
+export function copyFileToRemoteMachine(scriptFile: string, dest: string, scpConfig: any): Q.Promise<string> {
```
rm-dmiri, line 36: Nit: Seems like dest string is not used in the method, maybe remove it?
Also scpConfig is not used anymore. Or better: prefix with _ (_dest, _scpConfig) to keep the same signature.

Extensions/ArtifactEngineV2/IntegrationTests/proxyTests.ts
```diff
@@ -1,81 +1,51 @@
 import * as assert from 'assert';
-import * as http from 'http';
-import * as net from 'net';
 
 import * as engine from "../Engine"
 import * as providers from "../Providers"
 import { BasicCredentialHandler } from "../Providers/typed-rest-client/handlers/basiccreds";
 
-var packagejson = require('../package.json');
+var nock = require('nock');
```
rm-dmiri: Could you please check this concern?

> The switch from a real HTTP proxy server to nock means nock intercepts before the proxy layer runs. The test no longer validates:
> 
>  - proxy-authorization header
>  - user-agent header  
>  - actual HTTP CONNECT tunneling
> 
> Proxy auth regressions will now silently pass CI.

starkmsu (approved):

Extensions/Ansible/Tests/Tasks/Ansible/mockAnsibleUtils.ts
ivanduplenskikh: The mock should mirror the real signature, and we can't remove the unused arguments. I'd prefer to keep it as-is

Extensions/ArtifactEngineV2/IntegrationTests/proxyTests.ts
ivanduplenskikh: You are right.
Reverted the change and align it to the v1 where the proxy is 127.0.0.1:9011

Extensions/ArtifactEngineV2/IntegrationTests/proxyTests.ts
```diff
@@ -1,51 +1,81 @@
...
-            nock('http://redvstt-lab43:8080', { "encodedQueryParams": true })
-                .get('/job/ArtifactEngineJob/6/api/json')
-                .query({ "tree": "artifacts[*]" })
-                .basicAuth({ user: 'username', pass: 'password' })
-                .reply(200, {
-                    "artifacts": [
-                        { "displayPath": "file1.pdb", "fileName": "file1.pdb", "relativePath": "Extensions/ArtifactEngine/TestData/Jenkins/file1.pdb" },
-                    ]
-                });
+            // nock isn't working well with tunnel proxy so setting up custom server
```
rm-dmiri: maybe this is not needed anymore because we don't use nock anymore?

ivanduplenskikh: Good catch.
It was inherited by v1, so I've removed it.

manolerazvan (approved):

rm-dmiri (approved):

PR #1375, opened 2026-05-11
Fix proxy tunneling: cast port to Number and read IV from key file

v-schhabra:
**Description:**

Self-hosted agents behind authenticated proxies fail with v0.273.0 of the DownloadBuildArtifacts task when running on Agent v4.270.0+ (Node 24
runtime). Two independent bugs combine to cause the failure:

Bug 1 — String port in tunnel config: URL.parse().port returns a string (e.g. "8443"), but the tunnel library expects a number. In Node 24,
passing a string port causes the HTTP CONNECT handshake to silently hang until the 60-minute task timeout.

Bug 2 — IV mismatch in proxy credential decryption: azure-pipelines-task-lib v5.2.4+ encrypts secrets via _exposeTaskLibSecret() (in internal.js
) using AES-256-CTR with a random 32-byte key and a random 16-byte IV, stored in the key file as key_base64:iv_base64. The artifact-engine's
_readTaskLibSecrets() was not updated to read this IV — it used a zero IV via createDecipheriv, producing garbled proxy passwords and resulting
in 407 Proxy Authentication Required.

**Fixes:** 
Bug 1 — Port string → number (HttpClient.ts):

Bug 2 — Read IV from key file (webClientFactory.ts):

Bug 1 casts the string port to a number (Node 24 hangs on string ports). Bug 2 reads the key:iv from the key file instead of using a zero IV,
matching how _exposeTaskLibSecret() in task-lib encrypts proxy passwords.

Documentation changes required: N

Added unit tests: N 

Attached related issue: Y - https://portal.microsofticm.com/imp/v5/incidents/details/792049165/summary

Checklist:

- [x] Version was bumped - N/A, CI/build-tooling changes only (no extension/task/library version affected).
- [x] Checked that applied changes work as expected

### DIFF
--- a/Extensions/ArtifactEngineV2/Providers/typed-rest-client/HttpClient.ts
+++ b/Extensions/ArtifactEngineV2/Providers/typed-rest-client/HttpClient.ts
@@ -330,7 +330,7 @@ export class HttpClient {
                 proxy: {
                     proxyAuth: proxy.proxyAuth,
                     host: proxy.proxyUrl.hostname,
-                    port: proxy.proxyUrl.port
+                    port: Number(proxy.proxyUrl.port) || (proxy.proxyUrl.protocol === 'https:' ? 443 : 80)
                 },
             };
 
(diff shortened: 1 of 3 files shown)


rm-dmiri (approved):
It would be nice to add some unit tests to the changed logic or some basic happy paths if they are not covered already. If the fix is urgent, this can be done in a separate PR.

v-schhabra:
> It would be nice to add some unit tests to the changed logic or some basic happy paths if they are not covered already. If the fix is urgent, this can be done in a separate PR.

I will create a new PR by adding the unit tests as I need to merge the PR asap before the courtesy push created for 274 deployments.

PR #1381, opened 2026-05-14
Automatically update Test extension version in CI pipeline before publishing

v-gayatrij:
**Description**: 

### Problem

CI test pipelines for extensions are triggered with every PR commit. This requires the updated extensions in PR to be installed in the org where CI pipelines exist. Thus, requiring the extensions from the PR to be built and published as TEST(private) extensions to marketplace.

If the extension version from a PR is already published as test to marketplace during one of the previous commits, CI pipeline fails at publish step with below error.

```
error: Version number must increase each time an extension is published.  Extension: 
ms-vscs-rm.vss-services-bitbucket-test  Current version: 15.275.14  Updated version: 15.275.12
##[error]Failed to publish ms-vscs-rm.vss-services-bitbucket-test-15.275.12.vsix
```

### Solution

This PR adds a step to call existing BumpExtensionVersion.ps1 script to automatically update versions for updated extensions in a PR before attempting to publish them.

**Added unit tests:** N

**Attached related issue:** AB#2364503

**Checklist**:

### DIFF
--- a/.pipelines/1es-migration/azure-pipelines-integration.yml
+++ b/.pipelines/1es-migration/azure-pipelines-integration.yml
@@ -109,6 +109,27 @@ extends:
           displayName: 'Auto-detect changed extensions'
           workingDirectory: '$(Build.SourcesDirectory)'
 
+        - task: AzureCLI@2
+          displayName: 'Bump extension versions to exceed Marketplace'
+          inputs:
+            azureSubscription: '1es-extensions-publication-secure-service-connection'
+            scriptType: pscore
+            scriptLocation: inlineScript
+            inlineScript: |
+              $extensions = '$(DetectedExtensions)' -split ';'
+              foreach ($ext in $extensions) {
+                $ext = $ext.Trim()
+                if (-not $ext) { continue }
+
+                $manifestPath = Join-Path '$(Build.SourcesDirectory)' "Extensions/$ext/Src/vss-extension.json"
+                if (Test-Path $manifestPath) {
+                  Write-Host "--- Bumping version for: $ext ---"
+                  & '$(Build.SourcesDirectory)/scripts/BumpExtensionVersion.ps1' -ManifestPath $manifestPath
+                } else {
+                  Write-Host "##[warning]Manifest not found for '$ext' at $manifestPath — skipping version bump."
+                }
+              }
+
         - task: PowerShell@2
           displayName: 'Build test resources'
           inputs:
--- a/.pipelines/1es-migration/azure-pipelines-integration.yml
+++ b/.pipelines/1es-migration/azure-pipelines-integration.yml
@@ -111,11 +111,13 @@ extends:
 
         - task: AzureCLI@2
           displayName: 'Bump extension versions to exceed Marketplace'
+          condition: ne(variables['DetectedExtensions'], '')
           inputs:
             azureSubscription: '1es-extensions-publication-secure-service-connection'
             scriptType: pscore
             scriptLocation: inlineScript
             inlineScript: |
+              $ErrorActionPreference = 'Stop'
               $extensions = '$(DetectedExtensions)' -split ';'
               foreach ($ext in $extensions) {
                 $ext = $ext.Trim()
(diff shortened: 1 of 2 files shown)


.pipelines/1es-migration/azure-pipelines-integration.yml
```diff
@@ -109,6 +109,33 @@ extends:
...
+            inlineScript: |
+              $ErrorActionPreference = 'Stop'
+              $extensions = '$(DetectedExtensions)' -split ';'
+              foreach ($ext in $extensions) {
+                $ext = $ext.Trim()
+                if (-not $ext) { continue }
+
+                $manifestPath = Join-Path '$(Build.SourcesDirectory)' "Extensions/$ext/Src/vss-extension.json"
+                if (Test-Path $manifestPath) {
+                  Write-Host "--- Bumping version for: $ext ---"
```
nemanjarogic, line 128: Have you tested this with the local version which is higher than the one on Marketplace (no bumping needed)? I know that according to the logic in the script, version shouldn't be updated, but want to be sure we checked this.

.pipelines/1es-migration/azure-pipelines-integration.yml
```diff
@@ -109,6 +109,33 @@ extends:
...
+                  $scriptPath = Join-Path '$(Build.SourcesDirectory)' 'scripts/BumpExtensionVersion.ps1'
+                  & $scriptPath -ManifestPath $manifestPath
```
nemanjarogic, line 130: 1. Since pipeline changes are not yet merged, do you know how it will look like when the author forgets to bump the version, and this script is invoked? Who will be author of a new commit (which account)? Will that be stored permanently (visible from PR changes)?
2. If new commit is pushed to bump the version, doesn't that mean the current pipeline run, which is executing this logic, will be "irrelevant' because the new commit will trigger a new pipeline run with update version?

.pipelines/1es-migration/azure-pipelines-integration.yml
```diff
@@ -109,6 +109,33 @@ extends:
           displayName: 'Auto-detect changed extensions'
           workingDirectory: '$(Build.SourcesDirectory)'
 
+        - task: AzureCLI@2
+          displayName: 'Bump extension versions to exceed Marketplace'
```
nemanjarogic, line 113: This is not directly related to your changes, but it also helps to fix

`error: Version number must increase each time an extension is published`

I talked with Igor two days ago and we noticed the problem.
1. PR pipeline runs on commit push and publishes version X.Y.Z on Marketplace
2. PR is merged
3. PR author wants to publish extension on Marketplace (PROD version)
4. PR for publishing runs and it starts from the beginning. That means it will try to publish test and PROD version of extension to Marketplace with X.Y.Z version. However, it will fail when it try to publish test extension because X.Y.Z version is already taken and it needs at least X.Y.Z+1 version.

The easiest solution that comes to my mind is to either have:
1. The same 'Bump extension versions to exceed Marketplace' task, which you added here, in the pipeline for publishing PROD versions too (https://dev.azure.com/mseng/AzureDevOps/_build?definitionId=21050&_a=summary)
2. Introduce optional parameter for skipping test version publishing for https://dev.azure.com/mseng/AzureDevOps/_build?definitionId=21050&_a=summary. This would be a simple change. We already have '‎skipCiTesting‎' parameter there and need to apply the same logic for 'skipTestPublishingToMarketplace' (new parameter)

.pipelines/1es-migration/azure-pipelines-integration.yml
v-gayatrij: Yes, in cases where local version > Marketplace version, version will not be updated.
[This pipeline run](https://dev.azure.com/mseng/AzureDevOps/AzureDevOps%20Team/_build/results?buildId=31529154&view=logs&j=c4fd719a-9f8f-5382-8fd0-8a1f7c381c43&t=f0a88774-dabf-5a56-48ce-60bb6d1b4baa&l=234#:~:text=No%20bump%20needed%3A%20local%20(0.275.11)%20%3E%20Marketplace%20max%20(0.275.10)) confirms it. Ansible version was set to 0.275.11(greater than the one in marketplace) in [this test PR](https://github.com/microsoft/azure-pipelines-extensions/pull/1392) and BumpExtensionVersion.ps1 script skipped bumping it.

.pipelines/1es-migration/azure-pipelines-integration.yml
v-gayatrij: 1. There will not be a commit. Going through the BumpExtensionVersion.ps1, it seems to write the version directly inside build agent's local checkout i.e. inside $(Build.SourcesDirectory) - https://github.com/microsoft/azure-pipelines-extensions/blob/6fd1db73f883def2784d09a64afc36505079f021/scripts/BumpExtensionVersion.ps1#L168.
So the change will not be permanently stored.

2. Yes, current pipeline run executing this logic will become irrelevant once new commit triggers new pipeline run with updated version.

PR #1403, opened 2026-05-28
Add check to verify version update

v-gayatrij:
**Description**: Adds a version verification step to the CI pipeline that checks whether all affected extensions have their vss-extension.json version bumped above the current Marketplace version. If any extension version is not bumped, the pipeline fails and blocks the PR from merging. 

**Changes:** 

-  Added an AzureCLI@2 step in .pipelines/1es-migration/azure-pipelines-integration.yml after extension detection and before the build step, gated on PR builds only.
 - Added -VerifyOnly switch to scripts/BumpExtensionVersion.ps1 — exits with code 1 if the local version ≤ Marketplace version,without modifying any files.
 - Added `scripts/VerifyExtensionVersions.ps1` — accepts the list of detected extensions (from the `detectChangedExtensions` gulptask) and calls `BumpExtensionVersion.ps1 -VerifyOnly`  to validate if versions are updated for affected extensions.

 -   Updated `gulpfile.js` to exclude documentation/image files (`.md`, `.txt`, `.png`, `.jpg`, `.gif`) from shared infrastructure detection. Version update will not be required if any of these files are updated. Also extended list of common files in `SHARED_INFRA_PREFIXES ` array to include all the common files changing which may affect extensions.
 - Updated `README.md` — documents when and why version bumps are required, including shared infrastructure changes.

Flow:

 1. `gulp detectChangedExtensions` identifies which extensions have changes (uses git diff internally)
 2. If shared/common files are changed (e.g., `Extensions/Common/`, `package.json`, `gulpfile.js`, etc.), ALL publishable extensions are flagged — as common file changes can introduce regressions across all extensions.
 3. Documentation/image files in shared paths are excluded from this rule
 4. For each detected extension, verifies local manifest version > Marketplace max
 5. Fails with actionable error messages listing all extensions that need a version bump and the command to fix them 

**Added unit tests:** N

**Attached related issue:** (Y/N) <Please add link to related issue here>

**Checklist**:

### DIFF
--- a/.pipelines/1es-migration/azure-pipelines-integration.yml
+++ b/.pipelines/1es-migration/azure-pipelines-integration.yml
@@ -111,13 +111,13 @@ extends:
 
         - task: AzureCLI@2
           displayName: 'Verify extension versions are bumped'
-          condition: eq(variables['Build.Reason'], 'PullRequest')
+          condition: and(eq(variables['Build.Reason'], 'PullRequest'), ne(variables['DetectedExtensions'], ''))
           inputs:
             azureSubscription: '1es-extensions-publication-secure-service-connection'
             scriptType: pscore
             scriptLocation: scriptPath
             scriptPath: '$(Build.SourcesDirectory)/scripts/VerifyExtensionVersions.ps1'
-            arguments: '-SourceDirectory "$(Build.SourcesDirectory)"'
+            arguments: '-Extensions "$(DetectedExtensions)" -SourceDirectory "$(Build.SourcesDirectory)"'
 
         - task: PowerShell@2
           displayName: 'Build test resources'
--- a/.pipelines/1es-migration/azure-pipelines-integration.yml
+++ b/.pipelines/1es-migration/azure-pipelines-integration.yml
@@ -234,7 +234,7 @@ extends:
         dependsOn:
         - Build_And_Package_Test_Extension
         - Publish_Test_Extension
-        condition: and(not(failed('Publish_Test_Extension')), eq(dependencies.Build_And_Package_Test_Extension.outputs['BuildAndPackage.DetectExtensions.HasExtensionChanges'], 'true'))
+        condition: and(succeeded('Publish_Test_Extension'), eq(dependencies.Build_And_Package_Test_Extension.outputs['BuildAndPackage.DetectExtensions.HasExtensionChanges'], 'true'))
         variables:
           CI_TESTS_ORG_URL: 'https://dev.azure.com/canarytest'
           CI_TESTS_PROJECT: 'PipelineTasks'
--- a/.pipelines/1es-migration/azure-pipelines-integration.yml
+++ b/.pipelines/1es-migration/azure-pipelines-integration.yml
@@ -109,6 +109,15 @@ extends:
           displayName: 'Auto-detect changed extensions'
           workingDirectory: '$(Build.SourcesDirectory)'
 
+        - task: AzureCLI@2
+          displayName: 'Verify extension versions are bumped'
+          inputs:
+            azureSubscription: '1es-extensions-publication-secure-service-connection'
+            scriptType: pscore
+            scriptLocation: scriptPath
+            scriptPath: '$(Build.SourcesDirectory)/scripts/VerifyExtensionVersions.ps1'
+            arguments: '-SourceDirectory "$(Build.SourcesDirectory)"'
+
         - task: PowerShell@2
           displayName: 'Build test resources'
           inputs:
--- a/.pipelines/1es-migration/azure-pipelines-integration.yml
+++ b/.pipelines/1es-migration/azure-pipelines-integration.yml
@@ -111,6 +111,7 @@ extends:
 
         - task: AzureCLI@2
           displayName: 'Verify extension versions are bumped'
+          condition: eq(variables['Build.Reason'], 'PullRequest')
           inputs:
             azureSubscription: '1es-extensions-publication-secure-service-connection'
             scriptType: pscore
--- a/scripts/BumpExtensionVersion.ps1
+++ b/scripts/BumpExtensionVersion.ps1
@@ -23,7 +23,9 @@
 [CmdletBinding()]
 param(
     [Parameter(Mandatory)]
-    [string]$ManifestPath
+    [string]$ManifestPath,
+
+    [switch]$VerifyOnly
 )
 
 $ErrorActionPreference = 'Stop'
@@ -161,6 +163,11 @@ $newVersionStr = $localVersionStr
 
 if ($marketplaceVersion) {
     if ($localVersion -le $marketplaceVersion) {
+        if ($VerifyOnly) {
+            Write-Host "##[error]Version NOT bumped: local ($localVersion) <= Marketplace ($marketplaceVersion)"
+            exit 1
+        }
+
         $newVersionStr = "$($marketplaceVersion.Major).$($marketplaceVersion.Minor).$($marketplaceVersion.Build + 1)"
         Write-Host "Bumping   : $localVersion -> $newVersionStr  (local <= Marketplace max)"
 
--- a/scripts/BumpExtensionVersion.ps1
+++ b/scripts/BumpExtensionVersion.ps1
@@ -19,6 +19,10 @@
 
 .PARAMETER ManifestPath
     Path to the vss-extension.json file to check and update if necessary.
+
+.PARAMETER VerifyOnly
+    If specified, the script will only verify if the local version is greater than
+    the Marketplace version and will return an error if a bump is needed.
 #>
 [CmdletBinding()]
 param(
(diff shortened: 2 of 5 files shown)


scripts/BumpExtensionVersion.ps1
```diff
@@ -23,7 +23,9 @@
 [CmdletBinding()]
 param(
     [Parameter(Mandatory)]
-    [string]$ManifestPath
+    [string]$ManifestPath,
+
+    [switch]$VerifyOnly
```
nemanjarogic, line 32: Can you describe `VerifyOnly` param in the comment section above?

scripts/VerifyExtensionVersions.ps1
```diff
@@ -0,0 +1,97 @@
...
+}
+$ErrorActionPreference = 'Stop'
+
+if (-not $changedFiles) {
+    Write-Host "No changed files detected — skipping version verification."
+    exit 0
+}
+
+[array]$changedExtensions = $changedFiles |
+    Where-Object { $_ -match '^Extensions/([^/]+)/' } |
```
nemanjarogic: What happens when someone changes `Extensions/ArtifactEngineV2` or `Extensions/Common` directory? In `Extensions `directory we have more than just extensions.

scripts/VerifyExtensionVersions.ps1
```diff
@@ -0,0 +1,97 @@
...
+    ForEach-Object { $Matches[1] } |
+    Select-Object -Unique |
+    Where-Object { $_ -ne 'Common' }
+
+if ($changedExtensions.Count -eq 0) {
```
nemanjarogic: I think we should also introduce additional rule to be safe. If someone changes Extensions/Common or `package.json `in root directory (or any other common file) this could potentially broke all/some extensions. For such situations I would enforce version number update for all available extensions in this repository.

scripts/VerifyExtensionVersions.ps1
nemanjarogic: Ok, according to the code below, I believe it will be detected these directories don't contain `vss-extension.json `so nothing will happen.

scripts/VerifyExtensionVersions.ps1
nemanjarogic: Of course, we can lose this rule for some files which obviously can't affect extensions (.md, .txt, etc.)

scripts/VerifyExtensionVersions.ps1
```diff
@@ -0,0 +1,97 @@
...
+    Write-Host ""
+}
+
+# ---------------------------------------------------------------------------
+# Report result
+# ---------------------------------------------------------------------------
+if ($failures.Count -gt 0) {
+    Write-Host "##[error]Version not bumped for: $($failures -join ', ')"
+    Write-Host "##[error]Please update the 'version' field in vss-extension.json to exceed the Marketplace version."
+    Write-Host "##[error]Run: scripts/BumpExtensionVersion.ps1 -ManifestPath Extensions/<name>/Src/vss-extension.json"
```
nemanjarogic: You can actually give the exact command to solve this issue and more "user-friendly" command which hides scripts.
_____
Run `gulp build --syncVersions <ExtensionName1>,<ExtensionName2>`
For more details check: https://github.com/microsoft/azure-pipelines-extensions/blob/master/README.md#how-to-build
______
And you can collect what extensions have problem in the foreach loop above, so we don't even need to use placeholders for extension names.

.pipelines/1es-migration/azure-pipelines-integration.yml
nemanjarogic, line 107: What is `gulp detectChangedExtensions` doing?

scripts/VerifyExtensionVersions.ps1
v-gayatrij: Updated error log to display  gulp build --syncVersions command

.pipelines/1es-migration/azure-pipelines-integration.yml
v-gayatrij: gulp detectChangedExtensions was added to find all changed extensions by comparing with PR target branch and sets DetectedExtensions pipeline variable.
This already implements the logic to verify whether any common files were updated or not as suggested by you[ in this comment](https://github.com/microsoft/azure-pipelines-extensions/pull/1403#discussion_r3316534639). 
If  any common file changes are detected, this adds all extensions with vss-extension.json to DetectedExtensions variable.

I have updated VerifyExtensionVersions.ps1 to reuse the detectedExtensions variable set by this task - https://github.com/microsoft/azure-pipelines-extensions/commit/f8d5da5f0c8f7c73a666f457291d140a91fc768d#diff-1854ddcf0e3e26808e3762850cbfa0132126ef52d71794a94c42e4efe862d2b0L120

scripts/BumpExtensionVersion.ps1
v-gayatrij: Added description for VerifyOnly flag

scripts/VerifyExtensionVersions.ps1
v-gayatrij: Publishing and testing all extensions if any common files are detected was already done with the help of [detectChangedExtensions](https://github.com/microsoft/azure-pipelines-extensions/blob/c5f308b590af772143bf384880b41964b7b7df36/gulpfile.js#L998)

Added a check to exclude any .md/.txt/image files - https://github.com/microsoft/azure-pipelines-extensions/pull/1403/changes/04460b347f5d6b135a16e4e648dc548e35a8942d

scripts/VerifyExtensionVersions.ps1
v-gayatrij: yes, also detectChangedExtensions gulp task already handles this. It uses [discoverPublishableExtensions()](https://github.com/microsoft/azure-pipelines-extensions/blob/c5f308b590af772143bf384880b41964b7b7df36/gulpfile.js#L944) which only returns directories containing Src/vss-extension.json. Directories like ArtifactEngineV2, Common, and Terraform are filtered out since they don't have a manifest.


I have removed duplicate code for identifying changed extensions from VerifyExtensionVersions.ps1. Using [DetectedExtensions ](https://github.com/microsoft/azure-pipelines-extensions/commit/f8d5da5f0c8f7c73a666f457291d140a91fc768d#diff-1854ddcf0e3e26808e3762850cbfa0132126ef52d71794a94c42e4efe862d2b0L120)variable set by detectChangedExtensions gulp task instead

scripts/VerifyExtensionVersions.ps1
```diff
@@ -0,0 +1,73 @@
+<#
```
nemanjarogic, line 1: While this is good, can we do one more additional thing to have everything covered?

Task-level versions are not checked right now. Even though, this won't cause pipeline failure, we should change them when only something about task is changed (logic, dependency, etc.)? Do you think it should be good to introduce similar detection/warning mechanism for tasks within extension too? We can also do this through a separate PR. 

Tasks should be simpler, because we don't need to check Marketplace version, etc. I would just check whether it's greater than master version when we have changes **only inside task directory** (ignore even common files).

scripts/VerifyExtensionVersions.ps1
nemanjarogic: I think we need to extend list of common files, because some of these are not recognized right now. I believe we should have this as list of common files:

```js
const SHARED_INFRA_PREFIXES = [
    'Extensions/Common/',
    'Extensions/ArtifactEngine/',
    'Extensions/ArtifactEngineV2/',
    'common.json',
    'externals.json',
    'package.json',
    'package-lock.json',
    'package.js',
    'package-utils.js',
    'base.tsconfig.json',
    'tsconfig.json',
    'gulpfile.js',
    'definitions/',
    'TaskModules/',
    'scripts/',
    '.pipelines/',
    'ci/'
];
```

scripts/VerifyExtensionVersions.ps1
```diff
@@ -0,0 +1,73 @@
...
+    [Parameter(Mandatory)]
+    [string]$Extensions,
+
+    [Parameter(Mandatory)]
+    [string]$SourceDirectory
+)
+
+$ErrorActionPreference = 'Stop'
+
+[array]$extensionList = $Extensions -split ';' | ForEach-Object { $_.Trim() } | Where-Object { $_ }
```
nemanjarogic, line 29: Can you update two things:
1. PR description to be in sync with new logic (common files require version number update)
2. README.md - make it clear why and when we need to update version number for extensions (even when common file is changed because that can introduce regression).

scripts/VerifyExtensionVersions.ps1
v-gayatrij: Sure let me add warning for this in a follow-up PR

scripts/VerifyExtensionVersions.ps1
v-gayatrij: Updated description and readme file

scripts/VerifyExtensionVersions.ps1
v-gayatrij: updated SHARED_INFRA_PREFIXES

nemanjarogic (approved):

PR #1405, opened 2026-05-28
Refactor and cleanup of gulpfile.js

ivanduplenskikh:
**Description**:

Refactor and cleanup of `gulpfile.js`. No functional change to the build, package, or test pipelines — this is a maintainability pass intended to modernize the file and prepare it for stricter type checking. The only user-visible behavior change is the removal of the undocumented `--syncversions` (lowercase) flag alias; `--syncVersions` continues to work as before.

### Changes

**Dependencies**
- Removed the `yargs` dependency. All CLI argument parsing now goes through `minimist` (already imported), with an explicit option list: `suite`, `testAreaPath`, `publisher`, `extension`, `syncVersions` (strings) and `perf`, `e2e`, `runAllSuites` (booleans).
- `package.json` / `package-lock.json` updated accordingly (lockfile shrinks by ~250 lines from dropping `yargs` and its transitive deps).

**Syntax modernization**
- `var` → `const`/`let` throughout `gulpfile.js`.
- Function expressions converted to function declarations for helpers like `copyGroup`, `copyGroups`, `copyCommonModules`, `createVsixPackage`, `cacheArchiveFile`, `cacheNpmPackage`, `cacheNuGetV2Package`.
- `runMochaSuitesSequentially` is now `async`.
- Optional chaining used for `child.stdout?.pipe(...)` / `child.stderr?.pipe(...)`.
- Whitespace/formatting normalization.

**Type-checking readiness**
- Added JSDoc with param/return types on most helpers.
- Added `// @ts-ignore` at known dynamic-typing spots (xliff parsing, localization key indexing).

**Minor robustness / cleanups**
- `errorHandler` now logs the error message before exiting (previously silent).
- `compileUIExtensions` explicitly `return;`s when no `tsconfig` is found.
- Renamed shadowed `package` variable to `pkg` in the NuGet v2 caching loop.
- `fs.readFileSync` calls for `lib.json` and `vss-extension.json` now pass an explicit `'utf-8'` / `'utf8'` encoding.
- Inlined the single-caller `executeCommand` helper into `createVsixPackage`.

### Behavioral change to call out
- The `args.syncversions` (lowercase) fallback alias has been dropped — only `--syncVersions` is accepted now. All other CLI flag names (`--testAreaPath`, `--publisher`, `--extension`, `--suite`, `--perf`, `--e2e`, `--runAllSuites`) are unchanged.

### What did not change
- Gulp task graph and task names (`build`, `test`, `package`, `locCommon`, etc.).
- Build outputs and packaging flow.
- Test discovery / mocha invocation logic.
- All other CLI flag names and defaults.

### Diff stats
~232 insertions / ~415 deletions overall, dominated by lockfile shrinkage from removing `yargs`.

**Added unit tests:** N — refactor only, no behavior change beyond the dropped lowercase alias.

**Attached related issue:** N

**Checklist**:

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1083,7 +1083,7 @@ function createVsixPackage(extensionName) {
         shell.mkdir("-p", extnOutputPath);
 
         const packagingArgs = ["extension", "create", "--manifest-globs", "vss-extension.json", "--root", extnManifestPath, "--output-path", extnOutputPath];
-        const packagingResult = cp.spawnSync("tfx", packagingArgs, { stdio: "pipe" });
+        const packagingResult = cp.spawnSync("tfx", packagingArgs, { stdio: "pipe", shell: true });
         if (packagingResult.status !== 0) {
             const stderr = packagingResult.stderr ? packagingResult.stderr.toString() : "";
             throw new Error(`command failed: tfx ${packagingArgs.join(" ")}\n${stderr}`);
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1216,7 +1216,9 @@ function cacheNpmPackage(name, version) {
         const cmdline = '"' + npmPath + '" install ' + name + '@' + version + ' --userconfig ' + path.join(__dirname, ".npmrc");
         const result = cp.execSync(cmdline);
 
+        // @ts-ignore
         if (result.status > 0) {
+            // @ts-ignore
             throw new Error('npm failed with exit code ' + result.status);
         }
     } finally {
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1035,7 +1035,8 @@ gulp.task("detectChangedExtensions", function (done) {
 gulp.task("package", function (cb) {
     // use gulp package --extension=<Extension_Name> to package an individual package
     if (options.extension) {
-        createVsixPackage(options.extension); return;
+        createVsixPackage(options.extension);
+        return;
     }
 
     fs.readdirSync(_extnBuildRoot).filter(function (file) {
@@ -1070,11 +1071,14 @@ function createVsixPackage(extensionName) {
         del(extnOutputPath);
 
         if (options.publisher) {
-            const manifest = JSON.parse(fs.readFileSync(path.join(extnManifestPath, "vss-extension.json"), 'utf8'));
+            const extensionManifestPath = path.join(extnManifestPath, "vss-extension.json");
+            console.log(`🔁 Updating publisher in manifest for ${extensionManifestPath} to "${options.publisher}"`);
+            const manifest = JSON.parse(fs.readFileSync(extensionManifestPath, 'utf8'));
             manifest.publisher = options.publisher;
-            fs.writeFileSync(path.join(extnManifestPath, "vss-extension.json"), JSON.stringify(manifest));
+            fs.writeFileSync(extensionManifestPath, JSON.stringify(manifest));
         }
 
+        console.log(`📦 Packaging extension: ${extensionName}`);
         shell.mkdir("-p", extnOutputPath);
         const packagingCmd = "tfx extension create --manifest-globs vss-extension.json --root " + extnManifestPath + " --output-path " + extnOutputPath;
         shell.exec(packagingCmd, { silent: true }, (code) => {
@@ -1082,6 +1086,7 @@ function createVsixPackage(extensionName) {
                 console.error(`command failed: ${packagingCmd}\nManually execute to debug`);
             }
         });
+        console.log(`✅ Packaged extension: ${extensionName}`);
     }
 }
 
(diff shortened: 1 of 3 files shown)


gulpfile.js
```diff
@@ -998,44 +1044,53 @@
...
+        if (options.publisher) {
+            const manifest = JSON.parse(fs.readFileSync(path.join(extnManifestPath, "vss-extension.json"), 'utf8'));
+            manifest.publisher = options.publisher;
+            fs.writeFileSync(path.join(extnManifestPath, "vss-extension.json"), JSON.stringify(manifest));
         }
-    });
+
+        shell.mkdir("-p", extnOutputPath);
+        const packagingCmd = "tfx extension create --manifest-globs vss-extension.json --root " + extnManifestPath + " --output-path " + extnOutputPath;
+        shell.exec(packagingCmd, { silent: true }, (code) => {
```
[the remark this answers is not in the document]
ivanduplenskikh: It's not a production code, will be fixed in the next releases

gulpfile.js
```diff
@@ -998,44 +1046,59 @@ gulp.task("package", function(cb) {
...
-    shell.exec(cmd, {silent: true}, function(code, output) {
-        if (code != 0) {
-            console.error("command failed: " + cmd + "\nManually execute to debug");
-        } else {
-           callback();
+        console.log(`📦 Packaging extension: ${extensionName}`);
+        shell.mkdir("-p", extnOutputPath);
+
+        const packagingArgs = ["extension", "create", "--manifest-globs", "vss-extension.json", "--root", extnManifestPath, "--output-path", extnOutputPath];
+        const packagingResult = cp.spawnSync("tfx", packagingArgs, { stdio: "pipe" });
```
nemanjarogic: Can you double check whether this is true?

Image

gulpfile.js
```diff
@@ -998,44 +1046,59 @@ gulp.task("package", function(cb) {
...
+        if (packagingResult.status !== 0) {
+            const stderr = packagingResult.stderr ? packagingResult.stderr.toString() : "";
+            console.error(`command failed: tfx ${packagingArgs.join(" ")}\n${stderr}`);
```
nemanjarogic: We don't return from here or throw error. That means this will be reported as success and we will even log success message in the log (line below). Should we change this?

package.json
nemanjarogic: Since yargs is deleted do we need types?

gulpfile.js
```diff
@@ -734,9 +765,13 @@ function runMochaSuite(label, patterns, ignorePatterns) {
...
 
-// Run a list of suites sequentially (do NOT fast-fail; we want a complete CI
-// signal). Returns a Promise that rejects if any suite failed.
-function runMochaSuitesSequentially(suites) {
+/**
+ * Run a list of suites sequentially (do NOT fast-fail; we want a complete CI signal).
+ * @param {{ label: string, patterns: string[], ignorePatterns?: string[] }[]} suites - An array of suite definitions, each with a label for logging, an array of glob patterns to match test files, and an optional array of glob patterns to ignore.
+ * @returns {Promise<void>} - a Promise that resolves when all suites have completed, with an error if any suite failed.
+ */
+async function runMochaSuitesSequentially(suites) {
```
nemanjarogic, line 773: Should we use `await `where possible with async function?

gulpfile.js
```diff
@@ -507,9 +532,8 @@ gulp.task("updateTestIds", function(cb) {
     cb();
 });
 
-gulp.task("syncVersions", function(cb) {
-    var input = args.syncVersions || args.syncversions;
-    if (!input) {
+gulp.task("syncVersions", function (cb) {
+    if (!options.syncVersions) {
```
nemanjarogic, line 536: This was more intentional (to accept both options), but I am also fine with removing it because `syncVersions` is documented everywhere.

package.json
ivanduplenskikh: Removed

gulpfile.js
ivanduplenskikh: Agree. Now we throw an error

gulpfile.js
ivanduplenskikh: Yup, let's support the only one

gulpfile.js
ivanduplenskikh: Nope, since all calls are in return of gulp task's callback and gulp wraps a call into fn.Undertaker that applies and unwraps correctly [all Promise-based and thenable](https://github.com/gulpjs/undertaker/blob/2d95b5273d6a61fd4ca09376e91faae1045bbbe2/README.md?plain=1#L33) calls

gulpfile.js
ivanduplenskikh: Right, good catch. I've added `shell: true`

nemanjarogic (approved):

PR #1409, opened 2026-05-28
Migrate from fs-extra to native fs module

ivanduplenskikh:
**Description**: Removes the `fs-extra` dependency from the build tooling and replaces its usage with Node.js's built-in `node:fs` module.

Changes:
- `gulpfile.js`: Replace `fs-extra` import with `node:fs`; switch the previously destructured `rmSync` call sites to `fs.rmSync` (clean task and `createVsixPackage`).
- `package.js`: Replace `fs-extra` import with `node:fs`; switch `fs.copy(src, dest, cb)` to `fs.cp(src, dest, { recursive: true }, cb)` in `copyCommonModules`.
- `package.json` / `package-lock.json`: Drop `fs-extra` and `@types/fs-extra` devDependencies and their transitive entries (`jsonfile`, `universalify`).

Rationale: `node:fs` (Node 16.7+) provides `rmSync` and recursive `cp`, making `fs-extra` redundant for these build scripts. Reduces dependency surface and supply-chain footprint.

**Added unit tests:** N – build-tooling refactor only; no product code or runtime behavior changed.

**Attached related issue:** N

**Checklist**:

### DIFF
--- a/package.js
+++ b/package.js
@@ -1,7 +1,7 @@
 const cp = require('child_process');
 const path = require('path');
 
-const fs = require('fs-extra');
+const fs = require('node:fs');
 const gulp = require('gulp');
 const shell = require('shelljs');
 const through = require('through2');
@@ -112,7 +112,7 @@ function copyCommonModules(currentExtnRoot, commonDeps, commonSrc, extensionSour
                     const src = path.join(commonSrc, dep.module, "Src/");
                     const dest = path.join(targetPath, dep.dest);
                     shell.mkdir('-p', dest);
-                    fs.copy(src, dest, function (err) {
+                    fs.cp(src, dest, { recursive: true }, function (err) {
                         if (err) return console.error(err)
                     });
                 })
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,13 +1,12 @@
 // node built-ins
 const cp = require('node:child_process');
-const { rmSync } = require('node:fs');
+const fs = require('node:fs');
 const os = require('node:os');
 const path = require('node:path');
 const { parseArgs } = require('node:util');
 
 // build/test script
 const admZip = require('adm-zip');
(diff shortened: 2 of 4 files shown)


package.js
```diff
@@ -112,7 +112,7 @@ function copyCommonModules(currentExtnRoot, commonDeps, commonSrc, extensionSour
                     const src = path.join(commonSrc, dep.module, "Src/");
                     const dest = path.join(targetPath, dep.dest);
                     shell.mkdir('-p', dest);
-                    fs.copy(src, dest, function (err) {
+                    fs.cp(src, dest, { recursive: true }, function (err) {
```
nemanjarogic, line 115: It seems that fs.cp is async according to this: https://nodejs.org/api/fs.html#fscpsrc-dest-options-callback

In that case we should await the call before proceeding to next task dependency (this is in the loop).

There is also sync version fs.cpSync: https://nodejs.org/api/fs.html#fscpsyncsrc-dest-options

ivanduplenskikh: It was async as well - `fs.cp` has a callback. Not sure should we block the main thread with `cpSync` since we possibly can have numerous commonDeps.
But let's keep it as-is now. I plan to migrate `package.js` to `gulpfile.js` in future PRs and I'll make the proper refactoring

nemanjarogic: But we must think about possible bugs, and for example I am worried that we might see successful build with an incomplete artifact. The gulp task can finish and downstream stages (TS compile against the copied files, VSIX packaging, signing, publishing) can start before `fs.cp` has written all files. The .vsix then ships missing or partially-copied common modules, which will lead to the extension being broken at runtime.

nemanjarogic (approved):

package.js
ivanduplenskikh: Discussed privately to fix it in future PRs

PR #1416, opened 2026-06-03
Add TypeScript compilation task and improve error handling in gulpfile

ivanduplenskikh:
**Description**: Adds a new `tscBuildTasks` Gulp task that runs TypeScript compilation (`tsc -p`) as a build-time validation check for selected extensions, and improves the `errorHandler` message to handle undefined errors.

> [!WARNING]
> It's intermediate step tsc build implementation that will be enabled by default with full log output after all extensions will be onboarded and functioning as expected

- Reads the list of extensions/tasks to validate from `externals.json` under the `tsc-build-check` key.
- Recursively scans `ExtensionFolder` for `tsconfig.json` files (excluding `node_modules`) matching the allow-list and runs `tsc -p <configPath>` against each using the locally resolved `typescript/bin/tsc`.
- On failure, logs the failing config path and the exact `npx tsc -p ...` command to reproduce locally, then triggers `errorHandler` to fail the build.
- Wires the new `tscBuildTasks` step into the main `build` series between `compileNode` and `updateTestIds`.
- `errorHandler` now prints `Build failed with error: <error>` or `Build failed with an unknown error` instead of `Build failed undefined` when no error object is passed.

**Added unit tests:** N — build/gulp infrastructure change; covered by existing CI build runs.

**Attached related issue:** N

**Checklist**:
- [ ] Version was bumped — N/A (no extension/task/library version affected; build script change only).

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -60,7 +60,7 @@ const cultures = ['en-US', 'de-DE', 'es-ES', 'fr-FR', 'it-IT', 'ja-JP', 'ko-KR',
  * @returns {void}
  */
 function errorHandler(error) {
-    console.error(`Build failed ${error}`);
+    console.error(`Build failed ${error ? `with error: ${error}` : 'with an unknown error'}`);
     process.exit(1);
 }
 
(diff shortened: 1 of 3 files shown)


gulpfile.js
```diff
@@ -638,7 +638,33 @@ gulp.task("syncVersions", function (cb) {
...
+                console.error(`TypeScript compilation failed for ${configPath}: ${err.message}`);
+                console.error(`Execute manually the command:\nnpx tsc -p ${configPath}`);
+                cb();
+                errorHandler();
+            }
+        });
+    cb();
+});
+
+gulp.task("build", gulp.series("syncVersions", "compileNode", "tscBuildTasks", "updateTestIds"));
```
nemanjarogic, line 667: Can you share more information about how this should be used?
Can we also update main README.md and document when/how this should be used?

gulpfile.js
```diff
@@ -638,7 +638,33 @@ gulp.task("syncVersions", function (cb) {
...
+    }
+
+    const tscCliPath = require.resolve('typescript/bin/tsc');
+
+    fs.readdirSync(ExtensionFolder, { recursive: true, encoding: 'utf-8' })
+        .filter(x => buildCheckList.find((/** @type{string} */ check) => x.includes(check)) && path.parse(x).base == "tsconfig.json" && !x.includes("node_modules"))
+        .map(x => path.resolve(ExtensionFolder, x))
+        .forEach((configPath) => {
+            try {
+                cp.execFileSync(process.execPath, [tscCliPath, '-p', configPath], { stdio: 'ignore' });
```
nemanjarogic, line 656: Is there any reason to use stdio:ignore? I wonder could that sometimes hide necessary information. For example, in the case when err.message won't be enough to understand the problem. I am not sure whether this happens, but want to double check with you

ivanduplenskikh: To avoid the list of errors in CI mostly, if you have errors locally, it provides you the command to debug it, otherwise we have a mess of logs in a terminal

gulpfile.js
ivanduplenskikh: Readme has been updated

wawanawna (approved):

gulpfile.js
ivanduplenskikh: You were right, [I've added the condition](https://github.com/microsoft/azure-pipelines-extensions/pull/1424) that tracks isDebug value to show the output for better experience

PR #1421, opened 2026-06-04
Pipeline reliability updates: mandatory CI, optional test publish skip, and Marketplace validation polling

nemanjarogic:
**Description**: This PR improves extension publishing pipeline reliability and removes unnecessary post-merge test publishing.

What has changed:
1. CI testing is now mandatory in release and integration flows (all extensions are now covered by CI tests).
2. Added `skipTestExtensionPublishing` (default: `true`) in release pipeline because right now PR check pipeline will be always publish test extension and run CI tests. That means once PR is merged, we don't need to publish test version of extension again. In fact, attempt to do that will result with failure because version number will be the same and each publishing requires version number increase.
3. Added Marketplace validation polling with exponential backoff to handle long validation delays. Sometimes Marketplace validates extension upload for a few minutes and we don't want to show pipeline failure in such cases. It's better to wait and check what is the final status of upload. 
4. Renamed publishing script to reflect broader usage:
   - added `scripts/PublishMarketplaceExtensions.ps1`
   - removed `scripts/PublishTestExtensions.ps1`

**Documentation changes required:** Y

**Added unit tests:** N 

**Attached related issue:** AB#2425420

**Checklist**:


.pipelines/1es-migration/azure-pipelines.yml
```diff
@@ -224,124 +229,146 @@ extends:
...
+
+          - task: TfxInstaller@5
+            displayName: 'Install TFX'
+            inputs:
+              version: 'v0.21.1'
+
+          - task: 1ES.PublishAzureDevOpsExtension@1
+            name: PublishTestExtensionToMarketplace
+            displayName: 'Publish test extension'
+            continueOnError: true
```
ivanduplenskikh, line 304: This will appear green for the pipeline, but the task can fail silently.
Just to double-check, is it intended?

nemanjarogic: Depends on scenario and I would say we have three.

**1. Publish fails AND validation also fails**

Rescue reads the publish log → finds the message that validation of extension upload is not yet performed (Marketplace didn't confirm extension is uploaded) → enters polling loop.

tfx isvalid returns non-zero for all 5 attempts (~3.75 min) → rescue exits 1.
Job → Failed. Stage → Failed. 
Pipeline: red. 

**2. Publish fails, validation eventually succeeds**

Rescue reads the publish log → finds the message that validation of extension upload is not yet performed (Marketplace didn't confirm extension is uploaded) → enters polling loop.

An attempt of tfx isvalid returns 0 → rescue exits 0.
Publish step stays SucceededWithIssues (yellow), rescue step is green → job is SucceededWithIssues.
succeeded() is true for SucceededWithIssues, so Execute_CI_Tests → approval → PROD publish all proceed.
Pipeline: yellow. Intended rescue case.

**3. Publish fails for an unrelated reason (e.g. version already published, auth error, malformed package, network)**

Rescue reads the publish log → validation message not present (different error) → does not poll. 

Within seconds it exits 1 and echoes the last 40 lines of the publish log as ##[error], surfacing the real cause.
Job → Failed. Stage → Failed. Downstream stages skipped.
Pipeline: red, fast, with the real error visible in the rescue step itself.

ivanduplenskikh (approved):

PR #1423, opened 2026-06-04
Fix multiline PEM issue 

igortsoi:
**Description**: tl.setsecret raise error when multiline string is passed this shouldn't cause regression so all setsecret enclosed in try catch with tracing

**Added unit tests:** N

**Attached related issue:** N

**Checklist**:
- [X] Version was bumped - please check that version of the extension, task or library has been bumped.


Extensions/BitBucket/Src/Tasks/DownloadArtifactsBitbucket/downloadBitbucket.js
```diff
@@ -121,15 +121,15 @@ function getEndpointDetails(inputFieldName) {
         var token = getAuthParameter(bitbucketEndpoint, 'apitoken');
         var username = getAuthParameter(bitbucketEndpoint, 'email') || '';
         tl.debug('Using token authentication');
-        if (token) tl.setSecret(token);
+        try { if (token) tl.setSecret(token); } catch (e) { tl.debug('Failed to register secret for log masking: ' + e.message); }
```
ivanduplenskikh: Could you format these code lines as multiline to improve their readability?

Extensions/Ansible/Src/Tasks/Ansible/ansibleRemoteMachineInterface.ts
```diff
@@ -87,11 +87,23 @@ export class ansibleRemoteMachineInterface extends ansibleCommandLineInterface {
...
-        if (password) {
-            tl.setSecret(password);
+        try {
+            if (password) {
+                tl.setSecret(password);
+            }
+            if (privateKey) {
+                // PEM private keys are multi-line; setSecret() rejects multi-line values,
+                // so register each line individually to mask partial key material in logs.
+                for (const line of privateKey.split(/\r?\n/)) {
```
ivanduplenskikh, line 95: I'd add unit tests to cover the case

Extensions/Ansible/Src/Tasks/Ansible/ansibleTowerInterface.ts
```diff
@@ -37,8 +37,12 @@ export class ansibleTowerInterface extends ansibleInterface {
             this._hostname = tl.getEndpointUrl(connectedService, true);
 
             // Register credentials with the log masker so they are redacted in pipeline logs
-            if (this._password) {
-                tl.setSecret(this._password);
+            try {
+                if (this._password) {
```
ivanduplenskikh: All these `if` checks are excessive since the `setSecret` function already [verifies it](https://github.com/microsoft/azure-pipelines-task-lib/blob/a5e7ee0f29dbcfd3772a8882795777d60b0198bf/node/task.ts#L259)

Extensions/Ansible/Src/Tasks/Ansible/ansibleRemoteMachineInterface.ts
```diff
@@ -87,11 +87,23 @@ export class ansibleRemoteMachineInterface extends ansibleCommandLineInterface {
...
+                    if (line.length > 0) {
+                        tl.setSecret(line);
```
ivanduplenskikh, line 97: Should we trim the line?

And should we mask all content from PEM such as `preeb` and `posteb` that are described in [RFC1421](https://datatracker.ietf.org/doc/html/rfc1421#section-9) or we can skip some of them?

Extensions/Ansible/Src/Tasks/Ansible/ansibleRemoteMachineInterface.ts
```diff
@@ -87,11 +87,23 @@ export class ansibleRemoteMachineInterface extends ansibleCommandLineInterface {
...
+                    }
+                }
+            }
+        } catch (e) {
+            tl.debug('Failed to register secret for log masking: ' + e.message);
```
ivanduplenskikh: fyi: these debug lines won't show us the error output if SYSTEM_DEBUG variable is disabled and eventually all errors wrapped in tl.debug will be swallowed

Extensions/Ansible/Src/Tasks/Ansible/ansibleRemoteMachineInterface.ts
igortsoi: I will create CI test

Extensions/Ansible/Src/Tasks/Ansible/ansibleRemoteMachineInterface.ts
igortsoi: To not overcomplicated we mask everithing

Extensions/Ansible/Src/Tasks/Ansible/ansibleRemoteMachineInterface.ts
igortsoi: What would you suggest to use

Extensions/Ansible/Src/Tasks/Ansible/ansibleRemoteMachineInterface.ts
nemanjarogic: It's actually better to add CI tests first. That will allow you to test changes from your PR before you merge PR

Extensions/Ansible/Src/Tasks/Ansible/ansibleTowerInterface.ts
```diff
@@ -37,8 +37,10 @@ export class ansibleTowerInterface extends ansibleInterface {
             this._hostname = tl.getEndpointUrl(connectedService, true);
 
             // Register credentials with the log masker so they are redacted in pipeline logs
-            if (this._password) {
+            try {
                 tl.setSecret(this._password);
+            } catch (e) {
+                tl.debug('Failed to register secret for log masking: ' + e.message);
```
nemanjarogic: This will have effect of silent failure (we will have error, but it won't be reported as failure). Do we really want this? The same comment goes for other places with the same pattern.

igortsoi: I think setsecret shouldn't affect work of task

Extensions/Ansible/Src/Tasks/Ansible/ansibleRemoteMachineInterface.ts
ivanduplenskikh: It depends on what you need here: [we can show it as a warning or as an error](https://github.com/microsoft/azure-pipelines-task-lib/blob/a5e7ee0f29dbcfd3772a8882795777d60b0198bf/node/task.ts#L694-L696)

Extensions/Ansible/Src/Tasks/Ansible/ansibleRemoteMachineInterface.ts
ivanduplenskikh: Ok, but regarding the trim:

```js
// the output for this
tl.setSecret(' mypassword\u200b ');
// and for this will be different
tl.setsecret('mysecret');
```

Extensions/Ansible/Src/Tasks/Ansible/ansibleTowerInterface.ts
igortsoi: There can be some unexpected cases that will through some exception we can anticipate prevent all of them so but as this setSecret not related to business logic it should affect work of the task and break pipeline

Extensions/Ansible/Src/Tasks/Ansible/ansibleTowerInterface.ts
igortsoi: Thanks removed

Extensions/Ansible/Src/Tasks/Ansible/ansibleRemoteMachineInterface.ts
igortsoi: Not sure what you mean

Extensions/Ansible/Src/Tasks/Ansible/task.json
```diff
@@ -16,7 +16,7 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 0,
-        "Minor": 274,
+        "Minor": 275,
```
nemanjarogic, line 19: Should we also update task.json for other tasks? Or you didn't do that because you did through previous PR and extensions were not published?

Extensions/Ansible/Src/Tasks/Ansible/ansibleTowerInterface.ts
nemanjarogic: `tl.warning` seems more appropriate to me for this case. What do you think?

Extensions/Ansible/Src/Tasks/Ansible/task.json
igortsoi: Yes extension weren't published so the version bump is not needed

Extensions/Ansible/Src/Tasks/Ansible/ansibleTowerInterface.ts
igortsoi: It will display "Warning: Failed to register secret for log masking." on every run where this triggers, which would cause confusion and support tickets. The possible issue is not actionable there's nothing a user can do about it so there is no use of the warning

igortsoi: Changed to warning

nemanjarogic (approved):

ivanduplenskikh (approved):

nemanjarogic (approved):

PR #1427, opened 2026-06-05
Add TypeScript configuration and enhance JavaScript files with JSDoc comments in DownloadArtifactsTfsVersionControl

ivanduplenskikh:
**Description**:

Enables strict TypeScript checking (via `checkJs`) for the `DownloadArtifactsTfsVersionControl` task and annotates the JS files with JSDoc types to satisfy it.

Changes:
- **`tsconfig.json`** (new) — strict tsc config for this JS-only task:
  - `target: es6`, `module`/`moduleResolution: node16`, `esModuleInterop`, `skipLibCheck`, `noEmit`.
  - `checkJs: true`, `strict: true`, `useUnknownInCatchVariables: false`, `ignoreDeprecations: "6.0"`.
  - Explicit `files` list: `downloadTfsVersionControl.js`, `tfvcwrapper.js`.
- **`externals.json`** — add `DownloadArtifactsTfsVersionControl` to the `tsc-build-check` list so the new strict check runs in CI.
- **`downloadTfsVersionControl.js`** — make the file pass `checkJs` + `strict`:
  - Add JSDoc `@param`/`@returns`/`@type` annotations on every function and on the module-level `tl.getInput` calls (cast required inputs to `string`).
  - Switch `var` → `const`/`let`, arrow callbacks where applicable, consistent formatting.
  - Replace the `Q` promise library with native `Promise`s; `utilExec` now returns a real `Promise<{ code, output }>` and intermediate `.then` handlers return plain values instead of `Q(...)`. Remove the `Q` and unused `url` requires.
  - Null-check the result of `tl.getEndpointAuthorization` and throw the same decode error when missing.
  - `getAuthParameter` now returns `""` when the parameter is not found (instead of leaving `paramValue` undefined).
  - Replace `tl.exit(1)` with `process.exit(1)` in the error branches so the typed signature matches.
  - Rename shadowed `path` parameters to `folderPath` in `IsPathExists` / `ensurePathExist`.
  - Tighten `tfsEndpoint.Url` logging to guard against `undefined`.
- **`tfvcwrapper.js`** — add `// @ts-nocheck` at the top (legacy compiled TS output kept as-is) plus minor whitespace cleanup (tabs → spaces, trailing-space removal, single-line `listWorkspaces` body).
- **`task.json`** — indentation-only cleanup inside `dataSourceBindings`. No functional change.

**Added unit tests:** N — no behavior changes; the goal of this PR is to enable strict `tsc --checkJs` for the task and fix the resulting type errors.

**Attached related issue:** N

**Checklist**:

### DIFF
--- a/Extensions/ExternalTfs/Src/Tasks/DownloadArtifactsTfsVersionControl/downloadTfsVersionControl.js
+++ b/Extensions/ExternalTfs/Src/Tasks/DownloadArtifactsTfsVersionControl/downloadTfsVersionControl.js
@@ -1,67 +1,76 @@
-var shell = require('shelljs');
-var path = require('path');
-var tl = require('azure-pipelines-task-lib/task');
-var Q = require('q');
-var url = require('url');
-var tfvcwm = require('./tfvcwrapper');
+const path = require('path');
+const shell = require('shelljs');
 
-var projectId = tl.getInput("project");
-var repositoryId = tl.getInput("definition");
-var changesetId = tl.getInput("version");
-var downloadPath = tl.getInput("downloadPath");
-var tfsEndpoint = getEndpointDetails("connection");
+const tl = require('azure-pipelines-task-lib/task');
 
-console.log("project: " + projectId.toString());
-console.log("definition: " + repositoryId.toString());
-console.log("version: " + (!!changesetId ? changesetId.toString() : "Latest"));
-console.log("downloadPath: " + downloadPath.toString());
-console.log("tfsEndpoint:" + JSON.stringify(tfsEndpoint.Url));
+const tfvcwm = require('./tfvcwrapper');
 
-var tfvcw = new tfvcwm.TfvcWrapper();
-tfvcw.on('stdout', function (data) {
-    console.log(data.toString());
-});
-tfvcw.on('stderr', function (data) {
-    console.log(data.toString());
-});
-tfvcw.on('debug', function (data) {
-    console.log(data.toString());
-});
+const projectId = /** @type {string} */ (tl.getInput("project", true));
+const repositoryId = /** @type {string} */ (tl.getInput("definition", true));
+const changesetId = tl.getInput("version");
+const downloadPath = /** @type {string} */ (tl.getInput("downloadPath", true));
+const tfsEndpoint = getEndpointDetails("connection");
+
+console.log("project: " + String(projectId));
+console.log("definition: " + String(repositoryId));
+console.log("version: " + (!!changesetId ? String(changesetId) : "Latest"));
+console.log("downloadPath: " + String(downloadPath));
+console.log("tfsEndpoint:" + JSON.stringify(tfsEndpoint && tfsEndpoint.Url));
+
+/** @type {any} */
+const tfvcw = new tfvcwm.TfvcWrapper();
+
+tfvcw.on('stdout', (/** @type {any} */ data) => console.log(data.toString()));
+tfvcw.on('stderr', (/** @type {any} */ data) => console.log(data.toString()));
+tfvcw.on('debug', (/** @type {any} */ data) => console.log(data.toString()));
 tfvcw.setTfvcConnOptions({
     username: tfsEndpoint.Username,
     password: tfsEndpoint.Password,
     collection: tfsEndpoint.Url
 });
+
 getCode();
 
+/**
+ * @param {string} inputFieldName
+ */
 function getEndpointDetails(inputFieldName) {
-    var errorMessage = "Could not decode the External Tfs endpoint. Please ensure you are running the latest agent";
+    const errorMessage = "Could not decode the External Tfs endpoint. Please ensure you are running the latest agent";
     if (!tl.getEndpointUrl) {
         throw new Error(errorMessage);
     }
-    var externalTfsEndpoint = tl.getInput(inputFieldName);
+    const externalTfsEndpoint = tl.getInput(inputFieldName);
     if (!externalTfsEndpoint) {
         throw new Error(errorMessage);
     }
-    var hostUrl = tl.getEndpointUrl(externalTfsEndpoint, false);
+    const hostUrl = tl.getEndpointUrl(externalTfsEndpoint, false);
     if (!hostUrl) {
         throw new Error(errorMessage);
     }
-    var auth = tl.getEndpointAuthorization(externalTfsEndpoint, false);
+    const auth = tl.getEndpointAuthorization(externalTfsEndpoint, false);
+
+    if (!auth) {
+        throw new Error(errorMessage);
+    }
+
     if (auth.scheme != "UsernamePassword" && auth.scheme != "Token") {
         throw new Error("The authorization scheme " + auth.scheme + " is not supported for a External Tfs endpoint.");
     }
 
-    var hostUsername = ".";
-    var hostPassword = "";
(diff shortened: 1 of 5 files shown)


Extensions/ExternalTfs/Src/Tasks/DownloadArtifactsTfsVersionControl/tsconfig.json
```diff
@@ -0,0 +1,18 @@
...
+        "target": "es5",
+        "module": "node16",
+        "esModuleInterop": true,
+        "moduleResolution": "node16",
+        "skipLibCheck": true,
+        "noEmit": true,
+        "checkJs": true,
+        "useUnknownInCatchVariables": false,
+        "strict": true,
+        "ignoreDeprecations": "6.0"
```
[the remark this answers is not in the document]
ivanduplenskikh: Let's keep it since we plan to migrate it to ts v5 which supports this property

Extensions/ExternalTfs/Src/Tasks/DownloadArtifactsTfsVersionControl/tsconfig.json
```diff
@@ -0,0 +1,18 @@
...
+    },
+    "files": [
+        "downloadTfsVersionControl.js",
+        "tfvcwrapper.js"
+    ]
```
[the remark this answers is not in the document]
ivanduplenskikh: We'll mitigate it by refactor this file in future PRs

Extensions/ExternalTfs/Src/Tasks/DownloadArtifactsTfsVersionControl/downloadTfsVersionControl.js
```diff
@@ -152,45 +169,48 @@ function getCode() {
...
                     if (retCode === 0) {
                         console.log("Successfully added default mapping");
                     }
-                    return Q(retCode);
-                }, function(error) {
+                    return retCode;
+                }, function (/** @type {unknown} */ error) {
                     tl.error("Failed to add default mapping. " + error);
-                    tl.exit(1);
+                    process.exit(1);
```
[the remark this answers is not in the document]
ivanduplenskikh: Already discussed in https://github.com/microsoft/azure-pipelines-extensions/pull/1417#discussion_r3355684986

Extensions/ExternalTfs/Src/Tasks/DownloadArtifactsTfsVersionControl/downloadTfsVersionControl.js
```diff
@@ -152,45 +169,48 @@ function getCode() {
...
                     if (retCode === 0) {
                         console.log("Successfully synced workspace");
                     }
-                    return Q(retCode);
-                }, function(error) {
+                    return retCode;
+                }, function (/** @type {unknown} */ error) {
                     tl.error("Failed to sync workspace. " + error);
-                    tl.exit(1);
+                    process.exit(1);
```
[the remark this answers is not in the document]
ivanduplenskikh: Already discussed in https://github.com/microsoft/azure-pipelines-extensions/pull/1417#discussion_r3355684986

Extensions/ExternalTfs/Src/Tasks/DownloadArtifactsTfsVersionControl/downloadTfsVersionControl.js
```diff
@@ -1,149 +1,166 @@
...
                     if (retCode === 0) {
                         console.log("Successfully created workspace");
                     }
-                    return Q(retCode);
-                }, function(error) {
+                    return retCode;
+                }, function (/** @type {unknown} */ error) {
                     tl.error("Failed to Create a new Workspace. " + error);
-                    tl.exit(1);
+                    process.exit(1);
```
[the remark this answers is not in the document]
ivanduplenskikh: Already discussed https://github.com/microsoft/azure-pipelines-extensions/pull/1417#discussion_r3355684986

Extensions/ExternalTfs/Src/Tasks/DownloadArtifactsTfsVersionControl/tsconfig.json
```diff
@@ -0,0 +1,18 @@
...
+        "target": "ES6",
+        "module": "node16",
+        "esModuleInterop": true,
+        "moduleResolution": "node16",
+        "skipLibCheck": true,
+        "noEmit": true,
+        "checkJs": true,
+        "useUnknownInCatchVariables": false,
+        "strict": true,
+        "ignoreDeprecations": "6.0"
```
nemanjarogic: Can you explain what does this mean? What if we use typescript 5.5 instead of 4.9.5 (AI is suggesting me this might lead to errors)?

Extensions/ExternalTfs/Src/Tasks/DownloadArtifactsTfsVersionControl/downloadTfsVersionControl.js
```diff
@@ -1,149 +1,166 @@
...
         if (key.toLowerCase() === paramName.toLowerCase()) {
             keyName = key;
             return true;
         }
     });
-    paramValue = auth['parameters'][keyName];
-    return paramValue;
+
+    if (!keyName) {
+        return "";
```
nemanjarogic, line 97: Should we throw and report actual problem instead of having this silent error?

ivanduplenskikh: No, the change you suggested would change the current task flow.
I updated it to return the expected type, because all instances where `getAuthParameter` is called require a string

Extensions/ExternalTfs/Src/Tasks/DownloadArtifactsTfsVersionControl/tsconfig.json
ivanduplenskikh: Could you please provide which errors AI suggested?

ivanduplenskikh: This property just for swallowing errors which are related to outdated properties in tsconfig. You can find more details [here](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-6-0.html#preparing-for-typescript-70)

nemanjarogic: Here it is. 
image
But I see you removed this.

nemanjarogic (approved):

Extensions/ExternalTfs/Src/Tasks/DownloadArtifactsTfsVersionControl/tsconfig.json
ivanduplenskikh: I've removed this since it's not required anymore, because initially the problem occurred due to setting `es5` as the target, but I preferred to keep it as it is.]

ivanduplenskikh: > Here it is.

Got it, you referred to the error on build, yeah, it might be, but anyway, we removed it

PR #1439, opened 2026-06-17
Temporary hotfix to unblock DownloadArtifactsTfsVersionControl@15

nemanjarogic:
**Description**: 

Temporary hotfix to unblock `DownloadArtifactsTfsVersionControl@15` (External TFVC download), which currently aborts immediately for every customer with a `SyntaxError` at load time. The task still ships with the legacy **Node 6** execution handler, but its `shelljs` dependency was bumped to `^0.10.0`, which declares `engines.node >= 18` and uses JS syntax the Node 6 runtime cannot parse. The crash happens at `require('shelljs')` — before any auth, input validation, or `tf` invocation runs — so the task is 100% broken on current agents.

Failing agent log (from canary `ci-extensions/ExternalTfs/externaltfs_CI_test.yml`):
```
##[debug]Using node path: ...\externals\node\bin\node.exe
.../node_modules/shelljs/src/common.js:195
    );
    ^
SyntaxError: Unexpected token )
    at Object.<anonymous> (.../node_modules/shelljs/shell.js:9:14)
##[error]Exit code 1 returned from process: ...\externals\node\bin\node.exe
```

**Fix:** pin `shelljs` back to the last Node-6-compatible line (`^0.8.5`, `engines.node >= 4`). Also added a top-level `overrides` entry for `shelljs` because `azure-pipelines-task-lib@5.2.8` declares `shelljs: ^0.10.0` as a transitive dep — without the override, npm would install a nested 0.10.x copy under task-lib and the task would crash exactly the same way as soon as `require('azure-pipelines-task-lib/task')` runs. Lockfile was regenerated and now resolves a single `node_modules/shelljs@0.8.5` with no nested copy.

> **This is a temporary fix.** A follow-up PR will migrate the task off the EOL Node 6 handler to newer Node versions.

**Added unit tests:** N

**Attached related issue:** N

**Checklist**:
- [x] Version was bumped 
- [ ] Checked that applied changes work as expected — to verify by re-running the canary `externaltfs_CI_test.yml`

### DIFF
--- a/Extensions/ExternalTfs/Src/Tasks/DownloadArtifactsTfsVersionControl/package.json
+++ b/Extensions/ExternalTfs/Src/Tasks/DownloadArtifactsTfsVersionControl/package.json
@@ -5,10 +5,11 @@
     "@types/shelljs": "^0.8.17",
     "azure-devops-node-api": "11.2.0",
     "azure-pipelines-task-lib": "^5.2.8",
-    "shelljs": "^0.10.0"
+    "shelljs": "^0.8.5"
   },
   "overrides": {
     "underscore": "1.13.8",
-    "picomatch": "2.3.2"
+    "picomatch": "2.3.2",
+    "shelljs": "^0.8.5"
   }
 }
--- a/Extensions/ExternalTfs/Src/Tasks/DownloadArtifactsTfsVersionControl/package-lock.json
+++ b/Extensions/ExternalTfs/Src/Tasks/DownloadArtifactsTfsVersionControl/package-lock.json
@@ -9,7 +9,7 @@
         "@types/shelljs": "^0.8.17",
         "azure-devops-node-api": "11.2.0",
         "azure-pipelines-task-lib": "^5.2.8",
-        "shelljs": "^0.10.0"
+        "shelljs": "^0.8.5"
       }
     },
     "node_modules/@isaacs/cliui": {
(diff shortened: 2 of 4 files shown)


Extensions/ExternalTfs/Src/Tasks/DownloadArtifactsTfsVersionControl/package.json
```diff
@@ -5,10 +5,11 @@
...
     "azure-devops-node-api": "11.2.0",
     "azure-pipelines-task-lib": "^5.2.8",
-    "shelljs": "^0.10.0"
+    "shelljs": "^0.8.5"
   },
   "overrides": {
     "underscore": "1.13.8",
-    "picomatch": "2.3.2"
+    "picomatch": "2.3.2",
+    "shelljs": "^0.8.5"
```
ivanduplenskikh, line 13: Why is it necessary to override it if we've already set it in the **dependencies** section?

ivanduplenskikh:
Should we reintroduce shelljs correct one version and keep overrides for vulnerable subdependencies if we have small number of usages in the [source code](https://github.com/search?q=repo%3Amicrosoft%2Fazure-pipelines-extensions+%28%22shelljs%22+OR+%22shell%22%29+path%3AExtensions%2FExternalTfs%2FSrc%2FTasks%2FDownloadArtifactsTfsVersionControl%2F*.js&type=code)?
I'd change all those occurrences to native Nodejs methods in this case

Extensions/ExternalTfs/Src/Tasks/DownloadArtifactsTfsVersionControl/package.json
nemanjarogic: Yes, it's necessary. `dependencies `and `overrides `solve different problems.

`dependencies `only pins what our code directly requires (top of the tree).
Transitive deps resolve independently from whatever each parent package declares.
In our lockfile, you can see that `azure-pipelines-task-lib@5.2.8` declares its own `"shelljs": "^0.10.0"`. 

When task-lib internally calls into its own cp / mv / rmRF / mkdirP helpers, Node's module resolution walks up from task-lib's folder and finds the nested `0.10.0` first.

The overrides block forces every `shelljs` resolution anywhere in the tree to `^0.8.5`, including task-lib's transitive one.

ivanduplenskikh: Why do we use azure-pipelines-task-lib@5.2.8 which doesn't support Node6? Should we decrease the version of azure-pipelines-task-lib as well? Based on [engines](https://github.com/microsoft/azure-pipelines-task-lib/blob/350fdaca8e8ad707bf6974bd96bbb84a9847098e/node/package-lock.json#L38) in the package-lock.json file and usages in source code this library is valid only for >=Node18.
If we set engine-strict to true it fails. Should we respect those properties to avoid misunderstandings?

PR #1034, opened 2022-03-01
Moving Terraform extension to new repo

MOlausson:
This PR will delete the Terraform extension code and refer to the new repo where it will continue to be maintained. We also deprecate the Optimizely and Google Analytics extensions at this point.

PR #1440, opened 2026-06-18
Remove unused DownloadArtifactsTfsVersionControl task

nemanjarogic:
## Description

- **DownloadArtifactsTfsVersionControl** (task in `ms-vscs-rm.vss-services-externaltfs`) is removed because **telemetry shows no usage**. Abandoning support frees us from maintaining its Node 6 handler, custom TFVC wrapper, and tests.
- `DEPRECATION.md` is rewritten as a single ledger of removed extensions and tasks.

> Task GUID `F77A5E28-640E-4250-AA3A-174FA697A658` is retired and must not be reused.

Telemetry which shows no usage lowers the blast radius of this removal, but it doesn't make the Marketplace contract non-breaking. For example, we can still have classic releases with `DownloadArtifactsTfsVersionControl` references (just not run recently). Regardless, we want to remove this task and abandon support. However, in the case any breaking issue occurs we need to revert PR and publish extension again with task included.

## Checklist

-  Documentation changes required: Y
- Added unit tests: N
- Attached related issue: N
- Version was bumped: Y
- Checked that applied changes work as expected: Y


ivanduplenskikh (approved):

DEPRECATION.md
```diff
@@ -1,25 +1,31 @@
...
+| CircleCI artifacts for Release Pipeline | `ms-vscs-rm.vss-services-circleci-extension` | [#1420](https://github.com/microsoft/azure-pipelines-extensions/pull/1420) | 06-2026 |
+| Terraform | Source moved to [`microsoft/azure-pipelines-terraform`](https://github.com/microsoft/azure-pipelines-terraform) in [#1034](https://github.com/microsoft/azure-pipelines-extensions/pull/1034) | [#1440](https://github.com/microsoft/azure-pipelines-extensions/pull/1440) | 06-2026 |
 
+## Removed Tasks
 
+Individual tasks removed from extensions that otherwise remain published.
 
+| Task | Parent Extension | Removal PR | Removed Date |
+| --- | --- | --- | --- |
+| DownloadArtifactsTfsVersionControl (Download Artifacts - External TFVC) | TFS artifacts for Release Management (`ms-vscs-rm.vss-services-externaltfs`) | [#1440](https://github.com/microsoft/azure-pipelines-extensions/pull/1440) | 06-2026 |
```
manolerazvan, line 29: One thing I want to make sure we''ve thought through: even though telemetry shows no usage, pulling the published `ExternalTfvc` artifact-source-type and the task contribution is technically a breaking change for anyone who still has a classic release definition wired to an "External TFS Version Control" artifact. Once the new extension version rolls out, those definitions will stop being able to resolve/download that artifact.

I''m not against removing it — just want us to treat this as an intentional breaking retirement rather than a silent cleanup. Could we note the rollout/rollback plan (and any customer comms) in the PR description? "No telemetry" lowers the blast radius but doesn''t make the Marketplace contract non-breaking.

DEPRECATION.md
```diff
@@ -1,25 +1,31 @@
...
+| --- | --- | --- | --- |
+| Google Analytics | - | [#1128](https://github.com/microsoft/azure-pipelines-extensions/pull/1128) | 06-2021 |
+| Optimizely | - | [#1128](https://github.com/microsoft/azure-pipelines-extensions/pull/1128) | 06-2021 |
+| DataSources | - | [#1155](https://github.com/microsoft/azure-pipelines-extensions/pull/1155) | 05-2023 |
+| TwitterSentimentAnalysis | - | [#1187](https://github.com/microsoft/azure-pipelines-extensions/pull/1187) | 09-2023 |
+| Copy Stage variable to Release variable | `ms-vscs-rm.vss-services-copystagevariabletoreleasevariable` | [#1292](https://github.com/microsoft/azure-pipelines-extensions/pull/1292) | 09-2025 |
+| CircleCI artifacts for Release Pipeline | `ms-vscs-rm.vss-services-circleci-extension` | [#1420](https://github.com/microsoft/azure-pipelines-extensions/pull/1420) | 06-2026 |
+| Terraform | Source moved to [`microsoft/azure-pipelines-terraform`](https://github.com/microsoft/azure-pipelines-terraform) in [#1034](https://github.com/microsoft/azure-pipelines-extensions/pull/1034) | [#1440](https://github.com/microsoft/azure-pipelines-extensions/pull/1440) | 06-2026 |
```
manolerazvan, line 21: This Terraform row reads a little off to me. The "Marketplace Id" column is holding a "Source moved to…" sentence instead of an actual id, and the way it''s laid out it looks like #1440 removed Terraform in 06-2026 — but the source actually moved out in #1034 and all we''re doing here is deleting the leftover redirect README. Can we fix the column so it''s not misleading (maybe put the migration note somewhere it belongs, and keep the PR/date pointing at the real events)?

manolerazvan (commented):
Thanks for cleaning this up — the removal itself looks complete and I couldn't find any dangling references to the task, its GUID, or the Tfvc data-source bindings. Left a few inline notes on things I'd like us to sort out before merging.

Extensions/Ansible/Src/vss-extension.json
```diff
@@ -3,7 +3,7 @@
   "id": "vss-services-ansible",
   "name": "Ansible",
   "publisher": "ms-vscs-rm",
-  "version": "0.276.4",
+  "version": "0.277.0",
```
manolerazvan, line 6: This one caught my eye — Ansible (and BitBucket / IIS / ServiceNow / TeamCity) don''t contain the task we''re removing, so bumping their versions here just forces a republish of unrelated extensions. Unless the release tooling requires bumping everything in lockstep, I''d rather we drop these and keep the PR scoped to ExternalTfs. Same thought applies to deleting the Terraform README — feels like it belongs in its own change. If there''s a reason they all need to move together, a quick note in the description would clear it up.

DEPRECATION.md
nemanjarogic: Updated PR description.

DEPRECATION.md
nemanjarogic: Since PR is already merged, I updated notes in a new PR: https://github.com/microsoft/azure-pipelines-extensions/pull/1451

Extensions/Ansible/Src/vss-extension.json
nemanjarogic: The pipeline which triggers on every commit push forces us to update version number for each extension when common file changes. In this case, I changed `externals.json` and therefore had to update all extension numbers.

Regarding removal of Terraform through this PR, I agree that I could have done (PR is already merged due to Ivan's approval) that through separate PR, but since I was changing removal notes I just thought to clean up that too.

PR #1445, opened 2026-07-03
Add ServiceNow extension developer overview document

v-abhishera:
## Summary

Adds a comprehensive developer onboarding document for the ServiceNow Change Management extension and bumps the extension version.

## Changes

- **OVERVIEW.md** (new, 295 lines) — Internal developer documentation covering:
  - What the extension does and the problem it solves
  - How customers use it (YAML pipelines and classic release pipelines)
  - Extension components (tasks, service connections, data sources)
  - Key ServiceNow concepts (change requests, change types, success criteria)
  - Repository structure and key files
  - How to test and publish

- **vss-extension.json** — Version bump `4.278.1` → `4.278.2` 

## Why

The ServiceNow extension has no onboarding documentation for new developers, making it difficult for contributors to understand the architecture before making changes. This document provides a single entry point covering everything a developer needs to know.

## Testing

- No functional changes to the extension itself — documentation only + version bump.
- CI pipeline (`AzDev-ReleaseManagement-ServiceNow-CI-Test`) validates the extension still builds and passes end-to-end tests.

---


Extensions/ServiceNow/Src/OVERVIEW.md
```diff
@@ -0,0 +1,295 @@
...
+    environment:
+      name: Production          # This environment has a ServiceNow check configured
+    strategy:
+      runOnce:
+        deploy:
+          steps:
+          - script: echo "Deploying..."
+
+          # After deployment: mark the CR as closed
+          - task: ServiceNowUpdateChangeRequest@2
```
manolerazvan: The task name in this snippet is off. The registered task is `UpdateServiceNowChangeRequest`, but here it's written as `ServiceNowUpdateChangeRequest` (the `Update` and `ServiceNow` are swapped). Anyone who copy-pastes this into a pipeline will hit a "task not found" error, so it should be `- task: UpdateServiceNowChangeRequest@2`. Worth noting the components table earlier in the doc already uses the correct name, so the two currently disagree.

Extensions/ServiceNow/Src/OVERVIEW.md
```diff
@@ -0,0 +1,295 @@
...
+    ├── images/                     ← Screenshots used in readme.md
+    └── Tasks/
+        ├── CreateAndQueryChangeRequest/
+        │   ├── CreateAndQueryChangeRequestV0/  ← task.json + icon.png
+        │   ├── CreateAndQueryChangeRequestV1/  ← task.json + icon.png
+        │   └── CreateAndQueryChangeRequestV2/  ← task.json + icon.png (latest, recommended)
+        └── UpdateChangeRequest/
+            ├── UpdateChangeRequestV0/          ← task.json + icon.png
+            ├── UpdateChangeRequestV1/          ← task.json + icon.png
+            └── UpdateChangeRequestV2/          ← task.json + icon.png (latest, recommended)
```
manolerazvan, line 211: One thing that might save the next person some confusion: the folder is `UpdateChangeRequest` but the task it actually registers is `UpdateServiceNowChangeRequest` (that's the `name` field in task.json). The doc uses both names in different spots without calling out that one is the folder and the other is the registered task name. A quick one-line note here would make it clearer, and probably avoids the kind of mix-up that ended up in the YAML sample above.

Extensions/ServiceNow/Src/OVERVIEW.md
```diff
@@ -0,0 +1,295 @@
...
+| `scripts/VerifyExtensionVersions.ps1` | Verifies version was bumped in a PR. Called automatically by the PR check pipeline. |
+
+---
+
+## How to Test Your Changes
+
+CI tests run in the `canarytest` Azure DevOps organization against a real ServiceNow instance.
+
+| Pipeline | Purpose |
+|----------|---------||
```
manolerazvan: NIT: there's an extra pipe on this separator row (`|----------|---------||`), which can make the table render oddly in some markdown viewers. Dropping the trailing pipe fixes it.

Extensions/ServiceNow/Src/OVERVIEW.md
```diff
@@ -0,0 +1,295 @@
...
+---
+
+## Key ServiceNow Concepts
+
+### Change Request (CR)
+
+A record in ServiceNow tracking a planned change. Has a numeric state value:
+
+```
+New(-5) → Assess(-4) → Authorize(-3) → Scheduled(-2) → Implement(-1) → Review(0) → Closed(3)
```
manolerazvan, line 131: NIT: these state numbers are configurable per ServiceNow instance rather than fixed, so it'd be worth a short caveat that they're typical/example values. Not a blocker, just so nobody treats them as guaranteed.

manolerazvan (commented):
Took a pass through the new OVERVIEW.md and cross-checked the claims against the actual task.json / vss-extension.json. Overall it's accurate and genuinely useful (data source count, auth schemes, execution type, output vars, script and pipeline mappings all check out). One thing that would actually break if someone copies it, plus a couple of small cleanups below. One more minor note: the commit history bumps to 4.277.0 at one point, which is lower than master's 4.278.1; the final 4.278.2 is correct, so maybe just squash on merge.

Extensions/ServiceNow/Src/OVERVIEW.md
v-abhishera: Fixed — changed  ServiceNowUpdateChangeRequest@2  to  UpdateServiceNowChangeRequest@2  in the YAML snippet

Extensions/ServiceNow/Src/OVERVIEW.md
v-abhishera: Added a note clarifying that  UpdateChangeRequest  is the folder name while  UpdateServiceNowChangeRequest  is the registered task name in  task.json . This should prevent the same confusion going forward.

Extensions/ServiceNow/Src/OVERVIEW.md
v-abhishera: Fixed — removed the extra trailing pipe from the separator row.

Extensions/ServiceNow/Src/OVERVIEW.md
v-abhishera: Added a note clarifying these are typical default values and may be configured differently per ServiceNow instance.

manolerazvan (approved):

PR #1453, opened 2026-07-10
Add unit tests for Bitbucket Download Artifacts task

ivanduplenskikh:
**Description**: Adds a comprehensive unit test suite for the Bitbucket Download Artifacts task, and refactors the task source so it can be unit tested:
- Removed the `shelljs` dependency from `downloadBitbucket.js` and `sourcecontrolwrapper.js`, replacing `shell.rm`/`shell.cd`/`shell.which` with native `fs`, `process.chdir`, and `tl.which`/`tl.tool`.
- Removed unused `vso-node-api` imports.
- Made `definition`, `branch`, `version`, `downloadPath`, and `connection` task inputs required.
- Replaced the deprecated `.fail()` promise handler with `.catch()`, and fixed empty `catch {}` blocks (`catch (err) {}`) for broader engine/lint compatibility.

- Implement tests for various failure scenarios including:
  - Missing clone href
  - Empty clone links
  - Missing clone links
  - Missing SCM
  - Missing authentication scheme
  - Checkout failures
  - Cleanup exceptions
  - Clone failures
  - Malformed API responses
  - Missing authentication parameters (username, password, email)
  - Missing required inputs (branch, connection, definition, download path, version)
  - Unsupported authentication scheme
  - No endpoint authorization

- Add success scenarios for:
  - Case insensitive authentication parameters
  - Branch normalization
  - Cleanup with recursive option
  - Successful token and username/password authentication
  - Support for Mercurial SCM

- Introduce mock helpers for setting up test scenarios and environment variables.

**Added unit tests:** Y

**Attached related issue:** N

**Checklist**:

### DIFF
--- a/Extensions/BitBucket/Src/Tasks/DownloadArtifactsBitbucket/downloadBitbucket.js
+++ b/Extensions/BitBucket/Src/Tasks/DownloadArtifactsBitbucket/downloadBitbucket.js
@@ -16,12 +16,10 @@ var bitbucketEndpoint = getEndpointDetails("connection");
 
 removePathRecursive(downloadPath);
 
-
-var path = "/2.0/repositories/" + repositoryId;
 var options = {
     host: "api.bitbucket.org",
     method: "GET",
-    path: path
+    path: "/2.0/repositories/" + repositoryId
 };
 
 // Set authentication based on the endpoint type
--- a/Extensions/BitBucket/Src/Tasks/DownloadArtifactsBitbucket/downloadBitbucket.js
+++ b/Extensions/BitBucket/Src/Tasks/DownloadArtifactsBitbucket/downloadBitbucket.js
@@ -1,27 +1,20 @@
 var tl = require('azure-pipelines-task-lib/task');
+var fs = require('fs');
 var path = require('path');
-var webApim = require('vso-node-api/WebApi');
-var gitInterfaces = require('vso-node-api/interfaces/GitInterfaces');
 var Q = require('q');
 var url = require('url');
-var shell = require("shelljs");
 var scw = require('./sourcecontrolwrapper.js');
 var https = require("https");
 
 var BITBUCKET_API_TOKEN_AUTH_USERNAME = 'x-bitbucket-api-token-auth';
 
-var repositoryId = tl.getInput("definition");
-var branch = tl.getInput("branch");
-var commitId = tl.getInput("version");
-var downloadPath = tl.getInput("downloadPath");
+var repositoryId = tl.getInput("definition", true);
+var branch = tl.getInput("branch", true);
+var commitId = tl.getInput("version", true);
+var downloadPath = tl.getInput("downloadPath", true);
 var bitbucketEndpoint = getEndpointDetails("connection");
 
-shell.rm('-rf', downloadPath);
-var error = shell.error();
-if (error) {
-    tl.error(error);
-    tl.exit(1);
-}
+removePathRecursive(downloadPath);
 
 
 var path = "/2.0/repositories/" + repositoryId;
@@ -50,17 +43,17 @@ https.request(options, function (rs) {
         result = JSON.parse(response);
         tl.debug("result:" + JSON.stringify(result));
         var sch = new scw.SourceControlWrapper(result.scm);
-        
+
         sch.on('stdout', function (data) {
             console.log(data.toString());
         });
         sch.on('stderr', function (data) {
             console.log(data.toString());
         });
-        
+
         var remoteUrl = result.links.clone[0].href;
         tl.debug("Remote Url:" + remoteUrl);
-        
+
         // encodes projects and repo names with spaces
         var repoUrl = url.parse(remoteUrl);
         if (bitbucketEndpoint.Token) {
@@ -69,7 +62,7 @@ https.request(options, function (rs) {
         } else if (bitbucketEndpoint.Username && bitbucketEndpoint.Password) {
             repoUrl.auth = bitbucketEndpoint.Username + ':' + bitbucketEndpoint.Password;
         }
-        
+
         // if branch, we want to clone remote branch name to avoid tracking etc.. ('/refs/remotes/...')
         var ref;
         var brPre = 'refs/heads/';
@@ -79,7 +72,7 @@ https.request(options, function (rs) {
         else {
             ref = branch;
         }
-        
+
         var options = {
             creds: true,
             debugOutput: this.debugOutput
@@ -94,17 +87,14 @@ https.request(options, function (rs) {
 
         Q(0).then(function (code) {
             return sch.clone(repoUrl.format(repoUrl), false, downloadPath, options)
-                       .then(
-                function (code) {
-                    shell.cd(downloadPath);
+                .then(function (code) {
+                    process.chdir(downloadPath);
                     return sch.checkout(ref, options);
                 })
-                       .then(
-                function (code) {
+                .then(function (code) {
                     return sch.checkout(commitId);
                 })
-                       .fail(
-                function (error) {
+                .catch(function (error) {
                     tl.error(error);
                     tl.exit(1);
                 });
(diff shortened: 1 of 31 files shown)


Extensions/BitBucket/Src/Tasks/DownloadArtifactsBitbucket/downloadBitbucket.js
```diff
@@ -94,17 +87,14 @@ https.request(options, function (rs) {
 
         Q(0).then(function (code) {
             return sch.clone(repoUrl.format(repoUrl), false, downloadPath, options)
```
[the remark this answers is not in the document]
ivanduplenskikh: I'd prefer to keep the previous implementation. This PR should slightly touch the behavior.

Extensions/BitBucket/Src/Tasks/DownloadArtifactsBitbucket/downloadBitbucket.js
```diff
@@ -113,17 +103,17 @@ https.request(options, function (rs) {
 }).end();
 
 function getEndpointDetails(inputFieldName) {
-    var bitbucketEndpoint = tl.getInput(inputFieldName);
+    var bitbucketEndpoint = tl.getInput(inputFieldName, true);
     var auth = tl.getEndpointAuthorization(bitbucketEndpoint, false);
     var scheme = auth.scheme.toLowerCase().trim();
```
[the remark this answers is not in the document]
ivanduplenskikh: All validations will be implemented in a separate PR

Extensions/BitBucket/Src/Tasks/DownloadArtifactsBitbucket/sourcecontrolwrapper.js
```diff
@@ -53,19 +52,18 @@ var SourceControlWrapper = (function (_super) {
...
         var creds = this.username + ':' + this.password;
         var escapedCreds = encodeURIComponent(this.username) + ':' + encodeURIComponent(this.password);
         try {
             tl.setSecret(this.password);
             tl.setSecret(escapedCreds);
-        } catch {
+        } catch (err) {
             tl.warning('Failed to mask credentials for log redaction.');
         }
         tool.on('debug', function (message) {
```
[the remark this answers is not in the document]
ivanduplenskikh: Will be considered in a separate PR

Extensions/BitBucket/Tests/Tasks/DownloadArtifactsBitbucket/_suite.ts
```diff
@@ -0,0 +1,567 @@
...
+        sourceControlWrapperTl.warning = function (): void { return; };
+        sourceControlWrapperTl.tool = function (): any {
+            return {
+                on: function (name: string, cb: Function): void { handlers[name] = cb; },
+                arg: function (value: string): void { seenArgs.push(value); },
+                exec: function (ops: any): Promise<number> {
+                    seenExecOps = ops;
+                    handlers.debug('run https://u:p@bitbucket.org/org/repo.git and u%3Ap');
+                    handlers.stdout(new Buffer('stdout-line'));
+                    handlers.stderr(new Buffer('stderr-line'));
```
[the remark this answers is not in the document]
ivanduplenskikh: Fixed

nemanjarogic (approved):

Extensions/BitBucket/Tests/Tasks/DownloadArtifactsBitbucket/_suite.ts
```diff
@@ -0,0 +1,567 @@
...
+            const stderrEvents: string[] = [];
+            wrapper.on('stdout', function (data: any) { stdoutEvents.push(String(data)); });
+            wrapper.on('stderr', function (data: any) { stderrEvents.push(String(data)); });
+
+            const rc = await wrapper.exec(['clone', 'repo'], { debugOutput: true });
+
+            assert.strictEqual(rc, 0, 'exec should resolve with tool return code');
+            assert.strictEqual(JSON.stringify(seenArgs), JSON.stringify(['clone', 'repo']));
+            assert.ok(stdoutEvents.some(x => x.indexOf('[debug]') === 0), 'should emit debug output');
+            assert.ok(stdoutEvents.some(x => x.indexOf('...') >= 0), 'debug output should be masked');
```
manolerazvan, line 387: I don't think this test can pass as written. With username 'u' and password 'p', both creds and escapedCreds come out as 'u:p', but the debug message above uses the escaped form (u%3Ap), so there's nothing for the masking replace to match and no '...' ever gets produced. This line then expects '...' to be present, so it should fail. Since the agent already masks these values through setSecret, we could simply drop this test. If you'd rather keep it, use a password whose escaped form differs from the raw value, include both forms in the message, and assert that neither survives. It might also be worth confirming this suite is actually running in CI, because if it were, this assertion would be red.

Extensions/BitBucket/Src/Tasks/DownloadArtifactsBitbucket/downloadBitbucket.js
```diff
@@ -1,27 +1,20 @@
...
+var downloadPath = tl.getInput("downloadPath", true);
 var bitbucketEndpoint = getEndpointDetails("connection");
 
-shell.rm('-rf', downloadPath);
-var error = shell.error();
-if (error) {
-    tl.error(error);
-    tl.exit(1);
-}
+removePathRecursive(downloadPath);
```
manolerazvan, line 17: Could we use tl.rmRF(downloadPath) here, wrapped in a try/catch that calls tl.error and tl.exit(1) on failure, instead of the hand-rolled removePathRecursive? tl.rmRF keeps the Windows EBUSY/EPERM retry behavior and the dangling-symlink handling that the previous shell.rm -rf had. As it stands this runs at module load with no guard, so if cleanup fails it surfaces as an uncaught stack trace rather than a clean task failure like before.

Extensions/BitBucket/Src/Tasks/DownloadArtifactsBitbucket/downloadBitbucket.js
```diff
@@ -1,27 +1,20 @@
...
 
 
 var path = "/2.0/repositories/" + repositoryId;
```
manolerazvan: Small thing: this local variable shadows the path module that removePathRecursive uses. It works today only because the cleanup runs before this line, but it is a little fragile. Renaming it to something like apiPath would avoid a future surprise.

manolerazvan (commented):
Overall this looks good and the refactor reads cleanly. I left a few suggestions below, nothing blocking. One item for later: the task and extension versions were not bumped, which is fine while there is no release planned, just worth doing whenever this ships.

Extensions/BitBucket/Tests/Tasks/DownloadArtifactsBitbucket/_suite.ts
ivanduplenskikh: I double-checked and this does pass in [CI](https://dev.azure.com/mseng/PipelineTools/PipelineTools%20Team/_build/results?buildId=31764902&view=logs&j=f76d17c2-cc39-53e2-73c4-6a088ce9243c&t=d4d46235-5f46-54b6-3807-ee55045c3d50&l=155).
The mocked debug message actually embeds the raw `u:p` inside `https://u:p@bitbucket.org/...` , not just the trailing `u%3Ap`, so cred and escapedCreds (both 'u:p' here) [do match and get masked](https://github.com/microsoft/azure-pipelines-extensions/blame/59754cd94a4ad78ee3f4a4c15f3ac8acd2409e86/Extensions/BitBucket/Src/Tasks/DownloadArtifactsBitbucket/sourcecontrolwrapper.js#L64-L65).
Separately though, you're right that  repl = message.replace(escapedCreds, '...') overwrites instead of chaining from the first replace - that's a real latent bug which I'll fix it in a separate PR to make a decent refactoring separately

Extensions/BitBucket/Src/Tasks/DownloadArtifactsBitbucket/downloadBitbucket.js
ivanduplenskikh: Migrating to `tl.rmRF` that is defined in [the current dependency version](https://github.com/microsoft/azure-pipelines-extensions/blob/59754cd94a4ad78ee3f4a4c15f3ac8acd2409e86/Extensions/BitBucket/Src/Tasks/DownloadArtifactsBitbucket/package.json#L5) is not possible since there is a usage of `fs.rmSync` that was introduced in [Nodejs v14.14.0](https://nodejs.org/api/fs.html#fsrmsyncpath-options)

Extensions/BitBucket/Src/Tasks/DownloadArtifactsBitbucket/downloadBitbucket.js
ivanduplenskikh: Because this path is only used in the next by object, I pass the path variable value straight to it

manolerazvan (approved):

PR #1457, opened 2026-07-15
Use internal feed to install tfx and gulp

v-gayatrij:
**Description**: Fixes the extension publish pipeline failures caused by 1ES Network Isolation, which blocks access to external registries (e.g.  registry.npmjs.org ) and fails the global installs of  tfx-cli  and  gulp-cli . 

[Failing run](https://dev.azure.com/mseng/AzureDevOps/AzureDevOps%20Team/_build/results?buildId=31775380&view=logs&j=9936566d-898e-5ce3-0331-2c20a4700207&t=c12f02ff-8409-5f5f-a7a0-f37f67bac9f6&l=18)

Changes:
1. Move CLIs to root  devDependencies 
• Added  tfx-cli@^0.21.1  and  gulp-cli@^3.0.0  so they are installed locally from the internal CFS-upstream feed during  npm install , instead of via global installs /  TfxInstaller@5  which require external network access.
2. Pipeline updates ( azure-pipelines.yml  and  azure-pipelines-integration.yml )
• Removed the global  npm install -g tfx-cli  /  gulp-cli  steps and the  TfxInstaller@5  task.
• Added a step to prepend the local  node_modules/.bin  to  PATH  and set __tfxpath variable which is used by  PublishAzureDevOpsExtension task to resolve tfx.

**Documentation changes required:** (Y/N) N

**Added unit tests:** (Y/N) N

**Attached related issue:** N

**Checklist**:

### DIFF
--- a/.pipelines/1es-migration/azure-pipelines.yml
+++ b/.pipelines/1es-migration/azure-pipelines.yml
@@ -115,17 +115,10 @@ extends:
             command: install
             verbose: false
 
-        - task: Npm@1
-          displayName: 'Install TFX CLI (needed for publishing on Marketplace)'
-          inputs:
-            command: custom
-            customCommand: 'install -g tfx-cli --registry https://pkgs.dev.azure.com/mseng/PipelineTools/_packaging/PipelineTools_PublicPackages/npm/registry/'
-
-        - task: Npm@1
-          displayName: 'Install Gulp CLI'
-          inputs:
-            command: custom
-            customCommand: 'install -g gulp-cli --registry https://pkgs.dev.azure.com/mseng/PipelineTools/_packaging/PipelineTools_PublicPackages/npm/registry/'
+        - pwsh: |
+            $binPath = Join-Path "$(Build.SourcesDirectory)" "node_modules" ".bin"
+            Write-Host "##vso[task.prependpath]$binPath"
+          displayName: 'Add node_modules/.bin to PATH'
 
         - task: PowerShell@2
           displayName: 'Build PROD resources'
@@ -295,11 +288,26 @@ extends:
             inputs:
               versionSpec: '20.x'
 
-          - task: Npm@1
-            displayName: 'Install TFX CLI'
+          - checkout: self
+            displayName: 'Checkout repository'
+            clean: true
+            fetchTags: false
+
+          - task: NpmAuthenticate@0
+            displayName: 'Authenticate npm with internal CFS-upstream feed'
             inputs:
-              command: custom
-              customCommand: 'install -g tfx-cli --registry https://pkgs.dev.azure.com/mseng/PipelineTools/_packaging/PipelineTools_PublicPackages/npm/registry/'
+              workingFile: '$(Build.SourcesDirectory)/.npmrc'
+
+          - task: Npm@1
+            displayName: 'Install dependencies'
+            inputs:
+              command: install
+              verbose: false
+
+          - pwsh: |
+              $binPath = Join-Path "$(Build.SourcesDirectory)" "node_modules" ".bin"
+              Write-Host "##vso[task.prependpath]$binPath"
+            displayName: 'Add node_modules/.bin to PATH'
 
           - task: 1ES.PublishAzureDevOpsExtension@1
             name: PublishTestExtensionToMarketplace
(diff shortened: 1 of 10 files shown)


.pipelines/1es-migration/azure-pipelines.yml
```diff
@@ -408,10 +417,26 @@ extends:
           inputs:
             versionSpec: '20.x'
 
-        - task: TfxInstaller@5
-          displayName: 'Install TFX'
+        - checkout: self
+          displayName: 'Checkout repository'
```
nemanjarogic: Can you explain why do we need this?

.pipelines/1es-migration/azure-pipelines.yml
```diff
@@ -295,10 +288,26 @@ extends:
             inputs:
               versionSpec: '20.x'
 
-          - task: TfxInstaller@5
-            displayName: 'Install TFX'
+          - checkout: self
```
nemanjarogic: I assume this is not validated (no way before merge?) for PROD pipeline?

I believe the PATH-based approach works for the CI/integration pipeline but not for the PROD pipeline, because the two use different publish mechanisms:

Integration publishes via `PublishMarketplaceExtensions.ps1`, which calls `& tfx` directly → resolves from PATH, so `##vso[task.prependpath] …/node_modules/.bin` is enough.

PROD (jobs `PublishTestExtension` and `PublishProdExtension`) publishes via the `1ES.PublishAzureDevOpsExtension@1` task. Its shared [runTfx ](https://github.com/microsoft/azure-devops-extension-tasks/blob/main/BuildTasks/Common/Common.ts) resolves tfx only from the `__tfxpath` variable set by `TfxInstaller`. It has no PATH fallback and does not auto-install. With `TfxInstaller@5` removed and `__tfxpath` unset, I believe it would fail with: `Could not find tfx. To resolve, add the 'Use Node CLI for Azure DevOps' task to your pipeline before this task. `

Additionally, the [task repo's own test scripts](https://github.com/microsoft/azure-devops-extension-tasks/blob/main/Scripts/test-publishing.cmd#L0-L21) set `__TFXPATH=…\node_modules\.bin` and prepend `PATH ` meaning probably PATH alone isn't honored by the task.

If I am correct (validate this), the fix might be to keep the internal-feed approach: point the task at the locally installed tfx by setting `__tfxpath` alongside the PATH prepend, in both PROD publish jobs:

```
- pwsh: |
    $binPath = Join-Path "$(Build.SourcesDirectory)" "node_modules" ".bin"
    Write-Host "##vso[task.prependpath]$binPath"
    Write-Host "##vso[task.setvariable variable=__tfxpath]$binPath"
  displayName: 'Expose local tfx-cli to the publish task'
```

(_`_tfxpath` must be the directory containing the tfx launcher, i.e. `node_modules/.bin`.  keep the `prependpath` line too so the rescue script's & tfx still resolves.)

nemanjarogic: Another question, `TfxInstaller@5` needs to be removed because of network isolation? It's not available in internal feed?

.pipelines/1es-migration/azure-pipelines.yml
v-gayatrij: Checkout step was added as this job installs tfx and gulp from the repo's  package.json / .npmrc  via the internal feed (the previous  TfxInstaller  flow didn't touch the repo at all), so the job depends on the repository being checked out.

I have now removed it  as the job will anyhow do an implicit checkout as per [steps.checkout definition | Microsoft Learn](https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/steps-checkout?view=azure-pipelines) that will make these files available

.pipelines/1es-migration/azure-pipelines.yml
v-gayatrij: TfxInstaller@5 task was removed because it fetches the  tfx-cli  package from public npm registry. There doesn't seem to be an option with this task to dowlonad tfx from internal feed instead of public registry

v-gayatrij: Applied suggested fix - set __tfxpath in both PROD publish jobs. Could not validate this for PROD pipeline before merging due to branch control. Going through the task source code, PublishAzureDevOpsExtension task resolves tfx via __tfxpath variable as mentioned by you above.

nemanjarogic (approved):

PR #1463, opened 2026-07-20
Add OAuth2 support for bitbucket

v-snalawade:
**Description**:Add OAuth2 support for bitbucket

**Documentation changes required:** (Y/N) Y

**Added unit tests:** (Y/N) N

**Attached related issue:** (Y/N) [Feature](https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/2337780)

**Checklist**:
- [X] Version was bumped - please check that version of the extension, task or library has been bumped.
- [X] Checked that applied changes work as expected.


Extensions/BitBucket/Src/Tasks/DownloadArtifactsBitbucket/task.json
```diff
@@ -107,6 +107,14 @@
       "target": "downloadBitbucket.js",
       "argumentFormat": ""
     },
+    "Node24": {
```
nemanjarogic: Node 20 and 24 are duplicated and already exist in the list

Extensions/BitBucket/Src/Tasks/DownloadArtifactsBitbucket/downloadBitbucket.js
```diff
@@ -215,6 +266,13 @@ function getEndpointDetails(inputFieldName) {
             'Username': username
         };
     } else {
+        if (isRetireUsernamePswdEnabled) {
```
nemanjarogic, line 270: What is the plan for enabling retireusernamepswd? Asking because when enabled it will throw

Extensions/BitBucket/Tests/Tasks/DownloadArtifactsBitbucket/mockHelpers.ts
```diff
@@ -170,9 +170,28 @@ function registerFsMock(tr: tmrm.TaskMockRunner, cleanupFixture: boolean, cleanu
 
 function registerHttpsMock(tr: tmrm.TaskMockRunner, apiResponseText?: string): void {
     tr.registerMock('https', {
-        request: function (options: { [key: string]: string }, callback: Function) {
-            const auth = options.auth || '';
-            const authUser = auth.indexOf(':') >= 0 ? auth.split(':')[0] : auth;
+        request: function (options: { [key: string]: string | object }, callback: Function) {
```
nemanjarogic, line 188: This file was updated to decode the new Authorization header, and it even added a Bearer → [oauth2-bearer] branch. But:
- ScenarioOptions.scheme is still only `'Token' | 'UsernamePassword' | 'Unsupported'` and there is no `OAuth2.
- `setEndpointAuth()` has no oauth2 branch.
- There is no successOAuth2* scenario file.

Can we extend tests to test this feature?

Extensions/BitBucket/Src/Tasks/DownloadArtifactsBitbucket/task.json
nemanjarogic: One thing I noticed now. Is `Node20 `valid key? I believe we need to use `Node20_1`.

This won't cause runtime failure because the task also declares `Node16 `and `Node24`. The agent will just ignore the unknown `Node20 `key and run on `Node24` (or `Node16 `on old agents). However, the effect is that the intended `Node20 ` silently never engages.

v-snalawade: done, node changes are added by ivan and sergei [here](https://github.com/microsoft/azure-pipelines-extensions/commit/e457d2ab5e47152ff472e1f9a826f42a9b5ed514#diff-d2e909211c650faad645d6bac3cc2345b17da19c9ffcf9150606c6bdd3b2c6cb)

Extensions/BitBucket/Tests/Tasks/DownloadArtifactsBitbucket/mockHelpers.ts
v-snalawade: done

Extensions/BitBucket/Src/Tasks/DownloadArtifactsBitbucket/downloadBitbucket.js
v-snalawade: we will enable it when username/password scheme is fully deprecated from bitbucket

Extensions/BitBucket/Src/Tasks/DownloadArtifactsBitbucket/task.json
v-snalawade: done, it caused due to conflicts.

Extensions/BitBucket/Src/Tasks/DownloadArtifactsBitbucket/task.json
nemanjarogic: Can we change `Node20` to `Node20_1` so that we don't forget this later?

v-snalawade: done

Extensions/BitBucket/Src/Tasks/DownloadArtifactsBitbucket/downloadBitbucket.js
```diff
@@ -256,8 +314,8 @@ function getAuthParameter(endpoint, paramName) {
 
     const scheme = auth.scheme.toLowerCase().trim();
 
-    if (scheme !== 'usernamepassword' && scheme !== 'token') {
-        throw new Error('The authorization scheme ' + auth.scheme + ' is not supported for a bitbucket endpoint. Please use either "usernamepassword" or "token".');
+    if (scheme !== 'usernamepassword' && scheme !== 'token' && scheme !== 'oauth2') {
```
nemanjarogic: I believe the built-in Azure DevOps Bitbucket SC stores its OAuth grant as `Authorization.Scheme = "OAuth"`, not `"OAuth2"`. 

Evidence: the agent's own Bitbucket checkout (`BitbucketGitSourceProvider` → `GitSourceProvider.GetSourceAsync` in azure-pipelines-agent) switches on `endpoint.Authorization.Scheme` and handles `EndpointAuthorizationSchemes.OAuth` (value `"OAuth"`), reading the token from the `AccessToken` parameter; anything else hits the `default:` "Unsupport endpoint authorization schemes" branch. I can't see `OAuth2` scheme anywhere in the agent, and the constants are PascalCase-equal-to-value (`Token="Token"`, `UsernamePassword="UsernamePassword"`, `OAuth="OAuth"`), which is why the existing `scheme === 'token'` / `'usernamepassword'` checks already work.

If this is true, for a real OAuth connection (`"OAuth"` → `"oauth"`), the current check skips the new branch, falls through to the username/password path, and `getAuthParameter` then throws. The unit tests don't catch this because they mock `scheme: 'OAuth2'`.

Can you check this or confirm the current code works as expected? You confirm the actual value by enabling `System.Debug` on a pipeline using a real Bitbucket OAuth connection (or logging `tl.getEndpointAuthorizationScheme('connection')`)?

v-snalawade: You're right — the Azure DevOps built-in Bitbucket service connection uses "OAuth" (not "OAuth2"), matching the agent's EndpointAuthorizationSchemes.OAuth constant. The code has been updated to accept both [scheme === 'oauth'](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/7e7950df89/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and [scheme === 'oauth2'](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/7e7950df89/resources/app/out/vs/code/electron-browser/workbench/workbench.html) so real OAuth connections work correctly, while any custom connections that may have been configured with an OAuth2 label still function. The test mock has also been updated to use 'OAuth' to accurately reflect the real connection scheme.

nemanjarogic (approved):

PR #1468, opened 2026-07-21
Update adm-zip to 0.6.0 and bump affected extension versions

starkmsu:
**Description**: Update the `adm-zip` dependency to `0.6.0` across the entire repository, upgrade `azure-pipelines-task-lib` to `5.278.0`, modernize the Node execution handlers of the affected tasks, and bump the versions of all affected extensions, tasks and libraries.

### 1. adm-zip -> 0.6.0
- Root `package.json`: direct devDependency `adm-zip` `0.4.11` -> `0.6.0`.
- No `adm-zip` `overrides` are needed: `azure-pipelines-task-lib` `5.278.0` declares `adm-zip: ^0.6.0`, and `0.6.0` is the only `0.6.x` on the feed, so every transitive copy resolves to `0.6.0` on its own. The interim overrides added in earlier iterations were removed.

### 2. azure-pipelines-task-lib -> 5.278.0
- Bumped all 8 `azure-pipelines-task-lib` manifest entries to `5.278.0` (preserving the `^` range where it was already used).
- Regenerated all 8 `package-lock.json` files. Every lock now resolves `adm-zip` `0.6.0` and `azure-pipelines-task-lib` `5.278.0` only (no 0.4.x/0.5.x adm-zip remaining; `@types/adm-zip` is unrelated type defs).

### 3. Node execution handler modernization
`adm-zip` `0.6.0` declares `engines.node >=14.0`, so any sub-Node-14 execution handler could fail at runtime once selected. Replaced the old handlers and corrected `minimumAgentVersion`:
- Ansible: `Node10` -> `Node16`
- BitBucket `DownloadArtifactsBitbucket`: `Node` -> `Node16`
- TeamCity `DownloadTeamCityArtifacts`: `Node` -> `Node16`
- Set `minimumAgentVersion` to `2.206.1` (the Node16 handler floor per the azure-pipelines-task-lib agent matrix) on all affected tasks, including the two ExternalTfs tasks that already shipped Node16 but incorrectly declared `2.144.0` (the Node10 floor).
- `task.loc.json` files kept in sync with their `task.json` (version, handler, minimumAgentVersion).

### 4. Version bumps
Rule applied against the master baseline: if a version's Minor is already >= 278, bump Patch by 1; if Minor < 278, set Minor to 278 and Patch to 0.

- **Extensions (vss-extension.json):** Ansible `0.279.12`, BitBucket `15.279.18`, ExternalTfs `16.279.10`, IISWebAppDeploy `1.279.11`, ServiceNow `4.279.11`, TeamCity `15.279.14`
- **Tasks (task.json / task.loc.json):** Ansible `0.279.7`, DownloadArtifactsBitbucket `15.278.1`, DownloadArtifactsTfsGit `16.279.1`, DownloadExternalBuildArtifacts `16.279.1`, DownloadTeamCityArtifacts `15.279.2`
- **Libraries (package.json):** ArtifactEngine `1.278.0`, ArtifactEngineV2 `2.278.0`

> Note: IISWebAppDeploy and ServiceNow have no dependency changes; their extension versions were bumped at explicit request.

### 5. CI build fix: npm authentication for gulp sub-installs
Regenerating the lockfiles introduced packages not present in the root `package-lock.json`, which exposed a latent authentication gap in the `gulp build` step and caused it to fail with `npm error code E401`.

- Root cause: `runNpmInstall` (`gulpfile.js`) and `installDependencies` (`package.js`) ran per-extension / per-task `npm ci` with `--userconfig` pointing at the extension/task-local `.npmrc`, which carries only the registry URL and **no** auth token. Feed authentication (`NpmAuthenticate@0` in CI, `vsts-npm-auth` locally) is applied to the repository-root `.npmrc` only. These sub-installs previously succeeded solely because the earlier authenticated root `npm ci` pre-populated the shared npm cache; once uncached packages had to be fetched from the feed, they failed with E401.
- Fix: point both shared sub-install helpers at the repository-root `.npmrc` (matching the UI-contribution install and `cachePackage`, which already do this), so every extension/task `npm ci` is authenticated without per-extension pipeline wiring. Registry settings are identical across every `.npmrc` in the repo, so this only adds the credential. Verified: the `gulp build` step now passes in the PR validation build.

**Added unit tests:** (N) - dependency/version, task-manifest, and build-infrastructure change only.

**Attached related issue:** (N)

**Checklist**:

### DIFF
--- a/Extensions/Ansible/Src/Tasks/Ansible/task.json
+++ b/Extensions/Ansible/Src/Tasks/Ansible/task.json
@@ -20,6 +20,7 @@
         "Patch": 7
     },
     "demands": [],
+    "minimumAgentVersion": "2.206.1",
     "instanceNameFormat": "Run playbook",
     "groups": [
         {
@@ -324,7 +325,7 @@
         }
     ],
     "execution": {
-        "Node10": {
+        "Node16": {
             "target": "main.js",
             "argumentFormat": ""
         },
--- a/Extensions/Ansible/Src/Tasks/Ansible/task.json
+++ b/Extensions/Ansible/Src/Tasks/Ansible/task.json
@@ -17,7 +17,7 @@
     "version": {
         "Major": 0,
         "Minor": 279,
-        "Patch": 6
+        "Patch": 7
     },
     "demands": [],
     "instanceNameFormat": "Run playbook",
--- a/Extensions/ArtifactEngine/package.json
+++ b/Extensions/ArtifactEngine/package.json
@@ -16,13 +16,12 @@
     "email": "omeshp@microsoft.com"
   },
   "dependencies": {
-    "azure-pipelines-task-lib": "^5.2.8",
+    "azure-pipelines-task-lib": "^5.278.0",
     "handlebars": "4.7.9",
     "minimatch": "3.1.5",
     "tunnel": "0.0.4"
   },
   "overrides": {
-    "adm-zip": "0.6.0",
     "minimatch": "3.1.5",
     "picomatch": "2.3.2"
   },
--- a/Extensions/ArtifactEngine/package.json
+++ b/Extensions/ArtifactEngine/package.json
@@ -1,6 +1,6 @@
 {
   "name": "artifact-engine",
-  "version": "1.273.1",
+  "version": "1.278.0",
   "description": "Artifact Engine to download artifacts from jenkins, teamcity, vsts",
   "repository": {
     "type": "git",
@@ -22,6 +22,7 @@
     "tunnel": "0.0.4"
   },
   "overrides": {
+    "adm-zip": "0.6.0",
     "minimatch": "3.1.5",
     "picomatch": "2.3.2"
   },
--- a/Extensions/ArtifactEngineV2/package.json
+++ b/Extensions/ArtifactEngineV2/package.json
@@ -16,12 +16,11 @@
     "email": "omeshp@microsoft.com"
   },
   "dependencies": {
-    "azure-pipelines-task-lib": "^5.2.8",
+    "azure-pipelines-task-lib": "^5.278.0",
     "handlebars": "4.7.9",
     "typed-rest-client": "^3.0.0"
   },
   "overrides": {
-    "adm-zip": "0.6.0",
     "picomatch": "2.3.2"
   },
   "devDependencies": {
--- a/Extensions/ArtifactEngineV2/package.json
+++ b/Extensions/ArtifactEngineV2/package.json
@@ -1,6 +1,6 @@
 {
   "name": "artifact-engine",
-  "version": "2.276.1",
+  "version": "2.278.0",
   "description": "Artifact Engine to download artifacts from jenkins, teamcity, vsts",
   "repository": {
     "type": "git",
@@ -21,6 +21,7 @@
     "typed-rest-client": "^3.0.0"
   },
   "overrides": {
+    "adm-zip": "0.6.0",
     "picomatch": "2.3.2"
   },
   "devDependencies": {
--- a/Extensions/ExternalTfs/Src/Tasks/DownloadExternalBuildArtifacts/package.json
+++ b/Extensions/ExternalTfs/Src/Tasks/DownloadExternalBuildArtifacts/package.json
@@ -12,6 +12,7 @@
     "typescript": "5.1.6"
   },
   "overrides": {
+    "adm-zip": "0.6.0",
     "underscore": "1.13.8",
     "artifact-engine": {
       "handlebars": "4.7.9"
--- a/Extensions/ExternalTfs/Src/Tasks/DownloadExternalBuildArtifacts/package.json
+++ b/Extensions/ExternalTfs/Src/Tasks/DownloadExternalBuildArtifacts/package.json
@@ -5,14 +5,13 @@
     "@azure/msal-node": "^3.7.3",
     "artifact-engine": "^1.271.1",
     "azure-devops-node-api": "14.1.0",
-    "azure-pipelines-task-lib": "^5.2.8",
+    "azure-pipelines-task-lib": "^5.278.0",
     "azure-pipelines-tasks-azure-arm-rest": "^3.267.0"
   },
   "devDependencies": {
     "typescript": "5.1.6"
   },
   "overrides": {
-    "adm-zip": "0.6.0",
     "underscore": "1.13.8",
     "artifact-engine": {
       "handlebars": "4.7.9"
--- a/Extensions/ExternalTfs/Src/Tasks/DownloadExternalBuildArtifacts/task.json
+++ b/Extensions/ExternalTfs/Src/Tasks/DownloadExternalBuildArtifacts/task.json
@@ -12,7 +12,7 @@
     "Patch": 1
   },
   "demands": [],
-  "minimumAgentVersion": "2.144.0",
+  "minimumAgentVersion": "2.206.1",
   "inputs": [
     {
       "name": "connectionType",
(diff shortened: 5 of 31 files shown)


Extensions/ExternalTfs/Src/Tasks/DownloadExternalBuildArtifacts/package.json
```diff
@@ -12,6 +12,7 @@
     "typescript": "5.1.6"
   },
   "overrides": {
+    "adm-zip": "0.6.0",
```
nemanjarogic: Can we fix the actual problem where it exists, in `azure-pipelines-task-lib`? It's slower, but we did the same in the past.

package.json
```diff
@@ -25,7 +25,7 @@
     "@types/through2": "^2.0.41",
     "@types/validator": "^5.7.35",
     "@types/xml2js": "^0.4.14",
-    "adm-zip": "0.4.11",
+    "adm-zip": "0.6.0",
```
nemanjarogic, line 30: adm-zip 0.6.0 declares `engines.node >= 14`. This could potentially break extensions/task which are on lower version of node. This also means we might be in deadlock situation right now

nemanjarogic: Actually, this change might be fine for this file because adm-zip is used only from `gulpfile.js`. In other words, it used by devs or from pipelines where we have node >= 14. However, it's still higlhy risky for tasks with overrides

starkmsu: Previously used versions - 0.5.16 and 0.5.17 has a condition "node": ">=12.0", so it doesn't change anything in terms of support for Node 6 and Node 10 that we haven't abandoned yet.
Only this particular version - 0.4.11 - has a condition for node >= 3.0 so it was compatible with all supported node versions.
Since we plan to abandon Node 6 and Node 10 support anyway this year, I think this change is justified enough since it might push some users to upgrade node to higher versions on really old build agents.

Extensions/ExternalTfs/Src/Tasks/DownloadExternalBuildArtifacts/package.json
starkmsu: Good suggestion - let me update that lib first in a separate PR

starkmsu: Actually, that was already done via https://github.com/microsoft/azure-pipelines-task-lib/pull/1183

starkmsu: But that updated version of azure-pipelines-task-lib is not available yet on ADO feed, so we need to wait with updating the corresponding dependencies

package.json
nemanjarogic: > is justified enough since it might push some users to upgrade node to higher versions on really old build agents

Can we really say this? I agree we will abandon old node support soon, but until that happens what are our options? What will be our response if we get incident after we merge PR?

nemanjarogic (approved):

package.json
```diff
@@ -27,8 +27,8 @@
     "@types/through2": "^2.0.41",
     "@types/validator": "^5.7.35",
     "@types/xml2js": "^0.4.14",
-    "adm-zip": "0.4.11",
-    "azure-pipelines-task-lib": "5.2.8",
+    "adm-zip": "0.6.0",
+    "azure-pipelines-task-lib": "5.278.0",
```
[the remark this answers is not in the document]
starkmsu: With updated azure-pipelines-task-lib dependency we don't need anymore explicit override for adm-zip, it comes from this updated dependency now.

Extensions/ExternalTfs/Src/Tasks/DownloadExternalBuildArtifacts/package.json
```diff
@@ -5,7 +5,7 @@
     "@azure/msal-node": "^3.7.3",
     "artifact-engine": "^1.271.1",
     "azure-devops-node-api": "14.1.0",
-    "azure-pipelines-task-lib": "^5.2.8",
+    "azure-pipelines-task-lib": "^5.278.0",
     "azure-pipelines-tasks-azure-arm-rest": "^3.267.0"
```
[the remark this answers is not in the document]
starkmsu: With updated azure-pipelines-task-lib dependency we don't need any more explicit override for adm-zip, it comes from this updated dependency now. Having a version higher than 0.6.0 is fine since original security issue should be closed there.

Extensions/ArtifactEngine/package.json
```diff
@@ -16,7 +16,7 @@
     "email": "omeshp@microsoft.com"
   },
   "dependencies": {
-    "azure-pipelines-task-lib": "^5.2.8",
+    "azure-pipelines-task-lib": "^5.278.0",
     "handlebars": "4.7.9",
     "minimatch": "3.1.5",
     "tunnel": "0.0.4"
```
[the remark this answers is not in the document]
starkmsu: With updated azure-pipelines-task-lib dependency we don't need any more explicit override for adm-zip, it comes from this updated dependency now. Having a version higher than 0.6.0 is fine since original security issue should be closed there.

Extensions/ArtifactEngineV2/package.json
```diff
@@ -16,7 +16,7 @@
     "email": "omeshp@microsoft.com"
   },
   "dependencies": {
-    "azure-pipelines-task-lib": "^5.2.8",
+    "azure-pipelines-task-lib": "^5.278.0",
     "handlebars": "4.7.9",
     "typed-rest-client": "^3.0.0"
   },
```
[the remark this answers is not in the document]
starkmsu: With updated azure-pipelines-task-lib dependency we don't need any more explicit override for adm-zip, it comes from this updated dependency now. Having a version higher than 0.6.0 is fine since original security issue should be closed there.

Extensions/TeamCity/Src/Tasks/DownloadTeamCityArtifacts/package.json
```diff
@@ -3,7 +3,7 @@
   "main": "download.js",
   "dependencies": {
     "artifact-engine": "^1.271.1",
-    "azure-pipelines-task-lib": "^5.2.8",
+    "azure-pipelines-task-lib": "^5.278.0",
     "underscore": "^1.13.8"
```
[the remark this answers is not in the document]
starkmsu: With updated azure-pipelines-task-lib dependency we don't need any more explicit override for adm-zip, it comes from this updated dependency now. Having a version higher than 0.6.0 is fine since original security issue should be closed there.

Extensions/ExternalTfs/Src/Tasks/DownloadExternalBuildArtifacts/task.json
```diff
@@ -9,7 +9,7 @@
   "version": {
     "Major": 16,
     "Minor": 279,
-    "Patch": 0
+    "Patch": 1
   },
```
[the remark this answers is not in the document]
starkmsu: Fixed — bumped `task.loc.json` to 16.279.1 to match `task.json`, folded into the adm-zip commit. Both are now byte-identical on the version block.

Extensions/Ansible/Src/Tasks/Ansible/task.json
```diff
@@ -17,7 +17,7 @@
     "version": {
         "Major": 0,
         "Minor": 279,
-        "Patch": 6
+        "Patch": 7
     },
```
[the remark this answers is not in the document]
starkmsu: Here and in other tasks updated execution section to have Node16 as minimal supported version, minimumAgentVersion was also bumped to by in sync with Node16 handler.

package.json
starkmsu: Yes, we intentionally introduce this breaking change since all broken cases come from self-hosted agents. So clients are owning this and they will need to upgrade their Node anyway in several months. So for incidents related to this change our response will be to upgrade their Node to be at least Node16.

tarunramsinghani (approved):

PR #1485, opened 2026-07-31
Remove shelljs dependency and refactor file existence checks to use Node.js fs module

ivanduplenskikh:
**Description**: Remove the `shelljs` dependency and replace its usages with Node.js built-ins:
- `shell.exec` → `child_process.exec`
- `shell.test('-f')` / `shell.test('-d')` → `fs.existsSync` + `fs.lstatSync`
- `shell.which` → `tl.which`

Also converts `require()` calls for `fs` and `os` to TypeScript-style `import =` declarations.

**Added unit tests:** N

**Attached related issue:** N

**Checklist**:

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/Extensions/Ansible/Src/Tasks/Ansible/ansibleUtils.ts
+++ b/Extensions/Ansible/Src/Tasks/Ansible/ansibleUtils.ts
@@ -187,11 +187,11 @@ export function runCommandOnSameMachine(command: string, options: RemoteCommandO
 }
 
 export function testIfFileExist(filePath: string): boolean {
-    return fs.existsSync(filePath) && fs.lstatSync(filePath).isFile();
+    return fs.existsSync(filePath) && fs.statSync(filePath).isFile();
 }
 
 export function testIfDirectoryExist(directoryPath: string): boolean {
-    return fs.existsSync(directoryPath) && fs.lstatSync(directoryPath).isDirectory();
+    return fs.existsSync(directoryPath) && fs.statSync(directoryPath).isDirectory();
 }
 
 export function getAgentPlatform(): string {
@@ -199,7 +199,7 @@ export function getAgentPlatform(): string {
 }
 
 export function getShellWhich(moduleName: string): string | null {
-    return tl.which(moduleName, true);
+    return tl.which(moduleName, false);
 }
 
 export class WebRequest {
--- a/Extensions/Ansible/Src/Tasks/Ansible/ansibleUtils.ts
+++ b/Extensions/Ansible/Src/Tasks/Ansible/ansibleUtils.ts
@@ -1,5 +1,6 @@
-const fs = require('fs');
-var os = require('os');
+import cp = require('child_process');
+import fs = require('fs');
+import os = require('os');
 import http = require('http');
 import querystring = require('querystring');
 import util = require("util");
@@ -7,13 +8,16 @@ import util = require("util");
 import tl = require("azure-pipelines-task-lib/task");
 import httpClient = require('vso-node-api/HttpClient');
 import ssh = require('ssh2');
-import shell = require('shelljs');
 import SftpClient = require('ssh2-sftp-client');
 import Q = require("q");
 import crypto = require('crypto');
 
 var httpObj = new httpClient.HttpCallbackClient(tl.getVariable("AZURE_HTTP_USER_AGENT")!);
 const Ssh2Client = ssh.Client;
+
+const CP_EXEC_OPTIONS: cp.ExecOptions = {
+    maxBuffer: 20 * 1024 * 1024
+};
 
 export function _writeLine(str: string): void {
     process.stdout.write(str + os.EOL);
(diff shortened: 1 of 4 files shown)


Extensions/Ansible/Src/Tasks/Ansible/ansibleUtils.ts
```diff
@@ -162,10 +162,10 @@ export function runCommandOnSameMachine(command: string, options: RemoteCommandO
     var cmdToRun = command;
     tl.debug('cmdToRun = ' + cmdToRun);
 
-    shell.exec(cmdToRun, (err, _stdout, stderr) => {
+    cp.exec(cmdToRun, (err, _stdout, stderr) => {
```
nemanjarogic: Switching `shell.exec` with `cp.exec` has two implicit changes:

1. **Output is no longer shown.** shelljs `exec` echoes the child's stdout/stderr to the console by default. `cp.exec` doesn't echo, and this callback discards `_stdout`. Since the agent-local execution path routes through here (`ansibleCommandLineInterface._executeAnsiblePlaybook` → `executeCommand` → `runCommandOnSameMachine`), the `ansible-playbook` output and play recap will disappear from the pipeline log. 
2. **Large output now fails.** shelljs used `maxBuffer: 20 * 1024 * 1024` (`DEFAULT_MAXBUFFER_SIZE`); `cp.exec` defaults to 1 MB, so verbose ansible runs will hit `ENOBUFS`/"stdout maxBuffer length exceeded" and probably fail even when ansible succeeds.

Consider streaming, e.g. task-lib ToolRunner, which streams live, has no maxBuffer cap, and handles `failOnStdErr`:

    return tl.tool(tl.which('bash', true))
             .arg('-c')
             .arg(cmdToRun)
             .exec({ failOnStdErr: options.failOnStdErr } as any);

(or `cp.spawn(cmdToRun, { shell: true })` with stdout/stderr piped). 

Or pass a larger `maxBuffer` (if this is confirmed during testing) and forward `_stdout`/`stderr` to the log.

Extensions/Ansible/Src/Tasks/Ansible/ansibleUtils.ts
```diff
@@ -179,19 +179,19 @@ export function runCommandOnSameMachine(command: string, options: RemoteCommandO
...
+    return fs.existsSync(directoryPath) && fs.lstatSync(directoryPath).isDirectory();
 }
 
 export function getAgentPlatform(): string {
     return os.platform();
 }
 
 export function getShellWhich(moduleName: string): string | null {
-    return shell.which(moduleName);
+    return tl.which(moduleName, true);
```
nemanjarogic: `tl.which(moduleName, true)` throws when the tool isn't found, whereas the old `shell.which` returned `null`. The caller relies on a falsy return:

    if (!ansibleUtils.getShellWhich('ansible')) {
        throw tl.loc('AnisbleNotPresent');
    }

So when ansible is missing, this now throws a generic "Unable to locate executable file: 'ansible'…" and the friendly localized `AnisbleNotPresent` message becomes unreachable.

Extensions/Ansible/Src/Tasks/Ansible/ansibleUtils.ts
```diff
@@ -179,19 +179,19 @@ export function runCommandOnSameMachine(command: string, options: RemoteCommandO
 }
 
 export function testIfFileExist(filePath: string): boolean {
-    return shell.test('-f', filePath)
+    return fs.existsSync(filePath) && fs.lstatSync(filePath).isFile();
```
nemanjarogic: `shell.test('-f'/'-d', p)` follows symlinks - shelljs `test.js` uses `common.statFollowLinks(p)` (= `fs.statSync`). 

`fs.lstatSync` does not follow symlinks, so a playbook/inventory path that is a symlink to a file/dir now returns `false` and the task will throw `PlaybookNotPresent`/`InventoryNotFound` where it previously worked. Use `fs.statSync` to match the old behavior.

The same goes for change on line 186.

Also, is it possible to write tests to catch such a problem (if we already don't have them)?

Extensions/Ansible/Src/Tasks/Ansible/ansibleUtils.ts
ivanduplenskikh: Right, thank you.

I've set max buffer option to 20mb and output now the exec stdout under tl.debug

Extensions/Ansible/Src/Tasks/Ansible/ansibleUtils.ts
ivanduplenskikh: I've changed it to `statSync`, however there is a call of `createReadStream` in [ssh2-sftp-client library](https://github.com/theophilusx/ssh2-sftp-client/blob/c690045a5d05e40f86db6b7321c6e627071b6c4a/src/index.js#L675) that doesn't work with symlinks and should throw an error

Extensions/Ansible/Src/Tasks/Ansible/ansibleUtils.ts
ivanduplenskikh: Put false as a second argument for `tl.which` to avoid throws

nemanjarogic (approved):

nemanjarogic (approved):

PR #1496, opened 2026-08-05
Fix Ansible task output streaming and large-output failures

nemanjarogic:
## Summary

This change fixes a regression in the Ansible task where `Ansible@0` no longer displayed playbook output in pipeline logs and could fail with `ERR_CHILD_PROCESS_STDIO_MAXBUFFER` when playbook output was large.

The regression affected the local agent-machine execution path introduced in version `0.279.14` and `0.279.15` .

## Problem

The Ansible task previously used `shelljs.exec`, which streamed child process stdout and stderr directly to the task log. After the shelljs removal, the local command path was changed to `child_process.exec`.

`child_process.exec` is not a streaming API. It buffers stdout and stderr and only returns them to the callback after the process exits. In the merged implementation, stdout was not written to normal task output, so users could no longer see `ansible-playbook` output or the play recap in pipeline logs.

Because `exec` buffers output, large or verbose playbook runs could also exceed the configured `maxBuffer`, causing failures such as `ERR_CHILD_PROCESS_STDIO_MAXBUFFER`. Increasing the buffer only reduced the likelihood of failure; it did not remove the buffering problem.

## Fix

The local command execution path now uses `child_process.spawn` with `shell: true` instead of `child_process.exec`.

This restores streaming behavior:

- stdout is written directly to `process.stdout`, so Ansible output is visible in normal pipeline logs again.
- stderr is written directly to `process.stderr`, preserving visibility for warnings/errors.
- `failOnStdErr` behavior is still honored by tracking whether stderr was written and failing after the process exits when requested.
- Since output is streamed instead of buffered, there is no `maxBuffer` ceiling and large playbook output no longer triggers `ERR_CHILD_PROCESS_STDIO_MAXBUFFER`.

**Added unit tests:** Y

**Attached related issue:** N

**Checklist**:

### DIFF
--- a/Extensions/Ansible/Src/Tasks/Ansible/ansibleUtils.ts
+++ b/Extensions/Ansible/Src/Tasks/Ansible/ansibleUtils.ts
@@ -183,7 +183,9 @@ export function runCommandOnSameMachine(command: string, options: RemoteCommandO
             settled = true;
             tl.debug('code = ' + code + ', signal = ' + signal);
 
-            if (code && code != 0) {
+            if (signal !== null && signal !== undefined) {
+                reject(tl.loc('RemoteCmdExecutionErr'));
+            } else if (code && code != 0) {
                 reject(tl.loc('RemoteCmdNonZeroExitCode', cmdToRun, code));
             } else if (stdErrWritten === true && options.failOnStdErr === true) {
                 reject(tl.loc('RemoteCmdExecutionErr'));
--- a/Extensions/Ansible/Src/Tasks/Ansible/ansibleUtils.ts
+++ b/Extensions/Ansible/Src/Tasks/Ansible/ansibleUtils.ts
@@ -14,10 +14,6 @@ import * as httpClient from 'typed-rest-client/HttpClient';
 var httpObj = new httpClient.HttpClient(tl.getVariable("AZURE_HTTP_USER_AGENT")!);
 const Ssh2Client = ssh.Client;
 
-const CP_EXEC_OPTIONS: cp.ExecOptions = {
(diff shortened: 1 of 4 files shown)


tpavlovi (approved):

Extensions/Ansible/Src/Tasks/Ansible/ansibleUtils.ts
```diff
@@ -159,17 +155,40 @@ export function runCommandOnSameMachine(command: string, options: RemoteCommandO
...
+
+        child.on('close', (code, signal) => {
+            if (settled) {
+                return;
+            }
+
+            settled = true;
+            tl.debug('code = ' + code + ', signal = ' + signal);
+
+            if (code && code != 0) {
```
ivanduplenskikh: If we catch `SIGTERM` here (for example, [when agent kills the tree of processes](https://github.com/microsoft/azure-pipelines-agent/blob/891d3e106d49e7180d0e1b7d4d0000f2b3f27f0b/src/Agent.Sdk/ProcessInvoker.cs#L603)), code will be `null` and `signal` will be `SIGTERM`.
As a result, we could incorrectly reach `resolve('0')`, even though the process may not have produced any line and `stdErrWritten` may still be `false`.

An additional check would prevent this:

```js
if (signal !== null) {
    reject(tl.loc('RemoteCmdExecutionErr'));
} else if ...
```

nemanjarogic: Resolved.

ivanduplenskikh (approved):

PR #1497, opened 2026-08-06
Fix gulp task for syncing TEST version

v-gayatrij:
**Description**:

Add Installed Extensions API call to BumpExtensionVersion.ps1 script as a fallback to get the private TEST extension version.
The gulp build --syncVersions <Extensions> command returns  "TEST ver: not found." for some users. This could be due to lack of required permissions on the ms-vscs-rm publisher.

image

Changes:

- Read the version from the canarytest org's Extension Management API ( extmgmt.dev.azure.com/canarytest/_apis/extensionmanagement/installedextensions ) and match on  publisherId / extensionId . This reuses the same  az login  token master already uses for PROD. Using canarytest org here as the TEST extensions are already installed in this org which will auto update the installed extensions once published to marketplace.
- Bump the manifest version  for the six extensions (Ansible, BitBucket, ExternalTfs, IISWebAppDeploy, ServiceNow, TeamCity)

**Added unit tests:** N

**Attached related issue:** (Y/N) <Please add link to related issue here>

**Checklist**:

### DIFF
--- a/scripts/BumpExtensionVersion.ps1
+++ b/scripts/BumpExtensionVersion.ps1
@@ -35,10 +35,7 @@ param(
 $ErrorActionPreference = 'Stop'
 
 # ---------------------------------------------------------------------------
-# Obtain Azure CLI token (login is handled by the gulp task before this script runs).
-# The token is minted for the tenant that backs the canary org (Microsoft corp tenant);
-# a token from any other tenant is treated as anonymous by that org's API and returns
-# nothing for the private "-test" extension.
+# Obtain Azure CLI token (login is handled by the gulp task before this script runs)
 # ---------------------------------------------------------------------------
 $azToken = az account get-access-token --resource 499b84ac-1321-427f-aa17-267ca6975798 --tenant 72f988bf-86f1-41af-91ab-2d7cd011db47 --query accessToken -o tsv 2>$null
 if ($LASTEXITCODE -ne 0 -or -not $azToken) {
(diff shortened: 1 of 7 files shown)


scripts/BumpExtensionVersion.ps1
```diff
@@ -35,9 +35,12 @@ param(
 $ErrorActionPreference = 'Stop'
 
 # ---------------------------------------------------------------------------
-# Obtain Azure CLI token (login is handled by the gulp task before this script runs)
+# Obtain Azure CLI token (login is handled by the gulp task before this script runs).
```
nemanjarogic: Can you clarify PR description?
"`Fixes the gulp build --syncVersions   local version-bump flow so it can read the private -test  extension version when computing the next version. Previously the script only queried the public Marketplace Gallery, which did not include private extensions, so the test build always resolved as "TEST ver: not found.""`

Maybe I didn't understand something from earlier, but this is definitely not true. I just executed `gulp build --syncVersions Ansible` on master and it's working as it was always working. See results below.

Image

nemanjarogic: If the problem exist for someone, I assume it's related to permissions (on Marketplace?), but then we are switching this to potential problem with permissions on canarytest?

scripts/BumpExtensionVersion.ps1
```diff
@@ -130,6 +133,29 @@ function Get-MarketplaceVersion {
         Write-Warning "Failed to query Marketplace for '$FullExtensionId': $($_.Exception.Message)"
     }
 
+    # Private "-test" extensions are not returned by the public Gallery query above.
```
nemanjarogic: If your claim is true that test extensions are not returned by API above, why do we still call that API for test extensions?

scripts/BumpExtensionVersion.ps1
v-gayatrij: Apologies for the imprecise wording.

The reason public Gallery  extension query returns  private  -test  extension for you but not for me could be due to difference in authorization on ms-vscs-rm publisher.

For me and most other users, this returns  "TEST ver: not found." (screenshot in PR description)

scripts/BumpExtensionVersion.ps1
v-gayatrij: Removed public API call for test extensions and replaced it with [InstalledExtensions - ExtensionManagement](https://learn.microsoft.com/en-us/rest/api/azure/devops/extensionmanagement/installed-extensions/list?view=azure-devops-rest-7.1) API called using canarytest org inside Get-CanaryInstalledVersion()

scripts/BumpExtensionVersion.ps1
v-gayatrij: > If the problem exist for someone, I assume it's related to permissions (on Marketplace?), but then we are switching this to potential problem with permissions on canarytest?

Yes, this will still require permissions on canarytest to be able to get installed extensions but will relatively work for more users.

scripts/BumpExtensionVersion.ps1
```diff
@@ -133,14 +136,44 @@ function Get-MarketplaceVersion {
     return $null
 }
 
+# ---------------------------------------------------------------------------
+# Helper: read a private "-test" extension's version from the canary org where
+# it is installed (Extension Management API).
+# ---------------------------------------------------------------------------
+function Get-CanaryInstalledVersion {
```
nemanjarogic: The new canary Installed Extensions lookup breaks the PR version-verification gate.

In the newest current-head run, every canary query returns `401 Unauthorized`:

Image

Link to pipeline run from this PR: https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_build/results?buildId=31890381&view=logs&j=c4fd719a-9f8f-5382-8fd0-8a1f7c381c43&t=f0a88774-dabf-5a56-48ce-60bb6d1b4baa

The script catches those failures, returns `$null`, prints `TEST ver: not found`, compares only against PROD, and then reports `All extension versions are valid.` Therefore, the green pipeline does not prove the new lookup works; it proves the lookup failed open.

Before this change, the same gate successfully obtained the private TEST versions from the Gallery:

https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_build/results?buildId=31887961&view=logs&j=c4fd719a-9f8f-5382-8fd0-8a1f7c381c43&t=f0a88774-dabf-5a56-48ce-60bb6d1b4baa

Therefore, consider returning the working Gallery query as the primary source and use the canary API only as a fallback for local users who cannot read the private Gallery item.

v-gayatrij: Added canarytest API call as only a fallback when TEST version is not returned from Gallery.

nemanjarogic (approved):

PR #1507, opened 2026-08-14
Enhance TypeScript configuration for DownloadArtifactsTfsGit

ivanduplenskikh:
**Description**: Tightens the `DownloadArtifactsTfsGit` task's TypeScript checking and hardens its authentication-parameter handling against missing/malformed service connection data.

- Enabled a much stricter `tsconfig.json` for the task (`noImplicitAny`, `strictNullChecks`, `strictFunctionTypes`, `strictBindCallApply`, `strictPropertyInitialization`, `noImplicitThis`, `alwaysStrict`, `noUnusedLocals`, `noUnusedParameters`, `exactOptionalPropertyTypes`, `noImplicitReturns`, `noFallthroughCasesInSwitch`, `noUncheckedIndexedAccess`, `noImplicitOverride`, `noPropertyAccessFromIndexSignature`) to catch a wider class of type errors at build time.
- Made `getAuthParameter` null-safe: it now returns `null` instead of throwing when the endpoint's `auth` object or its `parameters` property is missing, rather than relying on an unchecked `auth['parameters']` lookup.
- Added an explicit guard in `getGitClientPromise` that fails fast with `"Username is not provided for the service connection."` when no username is available (in addition to the existing password guard), so a missing/incomplete auth parameters object produces a clear, actionable error instead of an obscure downstream failure.
- Updated JSDoc typedefs (`ConnectionDetails`, `AuthParameters`) to reflect the new nullable string types.
- Bumped the task/extension version per contribution guidelines.

**Added unit tests:** Y - added `failMissingAuthParameterObject.ts` (auth `parameters` object entirely absent) and `failMissingPassword.ts` (username present, password missing) fixtures, plus corresponding assertions in `_suite.ts` for both Node 16 and Node 20 runners.

**Attached related issue:** N

**Checklist**:

### DIFF
--- a/Extensions/ExternalTfs/Src/Tasks/DownloadArtifactsTfsGit/downloadTfGit.js
+++ b/Extensions/ExternalTfs/Src/Tasks/DownloadArtifactsTfsGit/downloadTfGit.js
@@ -20,6 +20,11 @@ const commitId = tl.getInput(isAdoConnectionType ? "versionAdo" : "version");
 const downloadPath = tl.getInput("downloadPath");
 validateInputs(serviceConnection, repositoryId, projectId, branch, commitId, downloadPath);
 
+/**
+ * @typedef {Object} AuthParameters
+ * @property {Record<string, any>} parameters - An object containing key-value pairs of authorization parameters.
+ */
+
 // @ts-ignore
 shell.rm('-rf', downloadPath);
 const error = shell.error();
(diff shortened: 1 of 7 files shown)


wawanawna (approved):
`allowUnreachableCode` and `allowUnusedLabels` are both `true`, which loosens checking -
  bit inconsistent with turning on `noUnusedLocals`/`noUnusedParameters`. Intentional, or
  leftover from a template?

ivanduplenskikh:
@wawanawna, right, it was a template typo.
Fixed that. In a separate PR I'll create a common tsconfig with **compilerOptions** and inherit it in all tasks' **tsconfig.json** files

wawanawna (approved):

PR #1508, opened 2026-08-14
Add support for syncing all publishable extensions in syncVersions command

ivanduplenskikh:
**Description**: This PR adds support for `--syncVersions all` when running the repo build flow. Previously, `--syncVersions` only accepted specific extension names or comma-separated extension lists, which required manual enumeration for multi-extension version bumps. The update extends the logic in `gulpfile.js` to detect the `all` case and resolve every publishable extension under `Extensions/` before invoking the version-bump script. The README is also updated to document the new usage.

**Documentation changes required:** Y

**Added unit tests:** N

**Attached related issue:** N

**Checklist**:


rm-dmiri (approved):
Check why build is failing, do we really need to bump versions in this case?

nemanjarogic (approved):

ivanduplenskikh:
@rm-dmiri, I don't think we need to bump versions here, since the change isn't directly or indirectly tied to the extensions or their build process

PR #1516, opened 2026-08-25
Harden Ansible task against OS command injection (CWE-78) behind feature flag

onetocny:
**Description**:

Fixes an OS command injection vulnerability (CWE-78) in the **Ansible** task. `ansibleCommandLineInterface._buildPlaybookExecutionCommand()` concatenated task inputs (`args`, `sudoUser`, `playbookPath`, inventory host list / inline content) into a single string that is executed via shelljs `shell.exec` (`/bin/sh -c`), so shell metacharacters in those inputs ran arbitrary commands on the agent.

Hardening applied, consistent with the SshV0 / CopyFilesOverSSHV0 / CMakeV1 fixes in `azure-pipelines-tasks`:

- `shellQuote` for the inventory (host list & file), playbook path, and sudo user. On the remote-machine (SSH) path, the `/tmp` inventory/playbook paths used by `-i` and the `rm -f` / `rm -rf` cleanup commands are hardened too; the raw path is retained for the shell-less SFTP/SCP copy.
- The free-form `args` field is neutralized via a quote-aware, single-pass transform `neutralizeAdditionalParameters` (in `argsSanitization.ts`): single-quoted content is left untouched, `$(...)` / backticks are neutralized even inside double quotes, control operators (`; | & < > ( ) #`) are escaped only outside quotes, unquoted `$VAR` / `${VAR}` and globs are preserved, and newlines are stripped. This replaces an earlier `shellSplit(...).map(neutralizeCommandSubstitution).join(' ')` approach, which corrupted JSON `--extra-vars` by dropping inner quotes.
- The inline inventory content is already hardened unconditionally on `master` via `shell-quote` (`quote([content])`, CodeQL SM03609 / Bug 2236220); that fix is preserved.
- The `shellEscaping` helpers are imported from the shared `azure-pipelines-tasks-utility-common/shellEscaping` package (no vendored copy).
- Applies to both the agent-machine and remote-machine (SSH) execution paths (shared base class).

The fix is gated behind dedicated feature flags, consumed via `tl.getPipelineFeature`, following the same enforce/shadow pattern as SshV0 / AzureCLI. The flags are keyed to this fix's own work item (Bug 2457936) so the Ansible rollout is staged independently of the shared `AZP_75787_*` sanitizer program:

- `AZP_2457936_ENABLE_NEW_LOGIC` — apply the hardening.
- `AZP_2457936_ENABLE_COLLECT` — emit telemetry only (shadow mode). Telemetry records a field when the hardened value differs from the legacy value (`legacyValue !== hardenedValue`) and tags the run `mode: live | shadow`.

With both flags off, the emitted command is byte-for-byte unchanged, so existing behavior and tests are unaffected.

**Added unit tests:** (Y)
- `testCommandInjectionHardeningForAgentMachine` — proves injection payloads (`; touch`, `$(whoami)`) are neutralized into inert literals when the flag is enabled.
- `argsSanitizationTests` — 27 direct unit tests for `neutralizeAdditionalParameters` (safe inputs round-trip, injection neutralized in/outside quotes, newlines stripped, idempotency).
- `testCommandInjectionHardeningForRemoteAgentSource` — remote + `file` + `agentMachine` with shell metacharacters in the inventory file name.
- `testExtraVarsJsonPreservedWhenHardeningEnabled` — flag-ON == legacy for JSON `--extra-vars`.
- `testSanitizationTelemetryShadowForAgentMachine` — shadow-mode telemetry signal.

**Attached related issue:** (Y) — Internal work item: mseng/AzureDevOps Bug 2457936.

**Checklist**:

- [x] Version was bumped - Ansible task `0.279.48` -> `0.279.50`.
- [x] Checked that applied changes work as expected - source type-checks clean and the full L0 suite passes (scenario + 27 unit + dependency tests) across Node 16/18/20, with legitimate args (`--extra-variables "pass=123" -v`, JSON `--extra-vars`) preserved.


onetocny:
Rebased onto latest master and resolved conflicts:
- **Version**: kept the version bump from #1515 and incremented patch -> `0.279.43`.
- **Inline inventory**: master already hardens the inline `echo` content unconditionally via `shell-quote` (`quote([content])` — CodeQL SM03609 / Bug 2236220). That line was kept **as-is** (not moved behind the feature flag) so the existing fix is not regressed. This PR adds hardening for the remaining, still-vulnerable inputs (`args`, `sudoUser`, `playbookPath`, inventory host list & file) behind the `AZP_75787_*` flags.
- Absorbed `master`'s TS refactor (const/style + `ansibleUtils.RemoteCommandOptions`).

Extensions/Ansible/Src/Tasks/Ansible/shellEscaping.ts
```diff
@@ -0,0 +1,89 @@
+// Shell-escaping helpers used to harden the ansible-playbook command against
```
onetocny: Import `azure-pipelines-tasks-utility-common` npm package and use `shellEscaping` from there rather than porting. See [readme](https://github.com/microsoft/azure-pipelines-extensions/blob/master/README.md) to properly build extension.

onetocny: Done in 7a73b109 — removed the vendored `shellEscaping.ts` and now import `shellQuote`/`neutralizeCommandSubstitution`/`shellSplit` from the `azure-pipelines-tasks-utility-common` package (`^3.272.0`; lockfile resolved 3.279.1, which includes the `shellEscaping` module added in 3.271.0). Added it to the task `package.json` dependencies so `gulp build` installs it per-task; `tsc -p` passes.

Extensions/Ansible/Src/Tasks/Ansible/ansibleCommandLineInterface.ts
```diff
@@ -158,29 +178,70 @@ export class ansibleCommandLineInterface extends ansibleInterface {
...
+
     protected _taskParameters: ansibleTaskParameters;
     protected _cleanupCmd: string[] = [];
     protected _remoteCmdOptions: ansibleUtils.RemoteCommandOptions;
     protected _playbookPath: string;
     protected _inventoryPath: string;
     private _sudoUser: string;
     private _additionalParams: string;
+    private _sanitizeActivate: boolean;
+    private _sanitizeAudit: boolean;
```
onetocny: Remove _sanitizeAudit and all the code behind it.

Extensions/Ansible/Src/Tasks/Ansible/ansibleCommandLineInterface.ts
```diff
@@ -158,29 +175,66 @@ export class ansibleCommandLineInterface extends ansibleInterface {
...
 
         if (this._sudoUser && this._sudoUser.trim()) {
-            cmd = cmd.concat('-b --become-user ' + this._sudoUser + ' ');
+            let sudoUser = this._applyHardening('sudoUser', this._sudoUser, this._sudoUser, shellQuote(this._sudoUser));
+            cmd = cmd.concat('-b --become-user ' + sudoUser + ' ');
         }
 
         if (this._additionalParams && this._additionalParams.trim()) {
-            cmd = cmd.concat(this._additionalParams);
+            let hardenedParams = shellSplit(this._additionalParams).map(neutralizeCommandSubstitution).join(' ');
```
onetocny: This change is not behind FF. If something happens in `shellSplit` or `neutralizeCommandSubstitution` then whole task will be influenced unconditionally. Same for `shellQuote` usage above.

onetocny: Good catch. I've made the hardening transform lazy: _applyHardening now takes a () => string producer, and shellQuote / shellSplit / 
eutralizeCommandSubstitution are only invoked when _sanitizeActivate (the FF) is on. With the flag off the method returns early with the legacy value and never touches those helpers, so the task is unaffected by them. Fixed in 5e7e4c2a.

Extensions/Ansible/Src/Tasks/Ansible/ansibleCommandLineInterface.ts
```diff
@@ -158,29 +175,71 @@ export class ansibleCommandLineInterface extends ansibleInterface {
...
+    private _applyHardening(fieldName: string, rawValue: string, legacyValue: string, harden: () => string): string {
+        if (!this._sanitizeActivate && !this._sanitizeTelemetry) {
+            return legacyValue;
+        }
+        if (rawValue && ansibleCommandLineInterface._shellMetaRegex.test(rawValue)) {
+            if (this._sanitizedFields.indexOf(fieldName) === -1) {
+                this._sanitizedFields.push(fieldName);
+            }
+        }
+        return this._sanitizeActivate ? harden() : legacyValue;
```
onetocny: Pass `rawValue` to `harden` instead of passing it to each lambda above.

Extensions/Ansible/Src/Tasks/Ansible/ansibleCommandLineInterface.ts
```diff
@@ -158,29 +175,71 @@ export class ansibleCommandLineInterface extends ansibleInterface {
...
 
         if (this._sudoUser && this._sudoUser.trim()) {
-            cmd = cmd.concat('-b --become-user ' + this._sudoUser + ' ');
+            let sudoUser = this._applyHardening('sudoUser', this._sudoUser, this._sudoUser, () => shellQuote(this._sudoUser));
+            cmd = cmd.concat('-b --become-user ' + sudoUser + ' ');
         }
 
         if (this._additionalParams && this._additionalParams.trim()) {
-            cmd = cmd.concat(this._additionalParams);
+            let additionalParams = this._applyHardening('additionalParameters', this._additionalParams, this._additionalParams, () => shellSplit(this._additionalParams).map(neutralizeCommandSubstitution).join(' '));
```
onetocny: Move `shellSplit(this._additionalParams).map(neutralizeCommandSubstitution).join(' ')` to separate function.

Extensions/Ansible/Src/Tasks/Ansible/ansibleCommandLineInterface.ts
```diff
@@ -52,7 +65,8 @@ export class ansibleCommandLineInterface extends ansibleInterface {
             const inventoryLocation = this._taskParameters.inventoryType;
 
             if (inventoryLocation == 'file') {
-                this._inventoryPath = this._taskParameters.inventoryFilePath;
+                let inventoryFilePath = this._taskParameters.inventoryFilePath;
```
onetocny: Use `const` whenever it is possible.

Extensions/Ansible/Src/Tasks/Ansible/ansibleCommandLineInterface.ts
onetocny: Done in 37f5a7a7 — _applyHardening now takes harden: (value: string) => string and calls harden(rawValue) internally. Call sites pass the transform directly (shellQuote, _neutralizeAdditionalParameters) with no capturing lambda.

Extensions/Ansible/Src/Tasks/Ansible/ansibleCommandLineInterface.ts
onetocny: Done in 37f5a7a7 — extracted into a private static helper _neutralizeAdditionalParameters(value).

Extensions/Ansible/Src/Tasks/Ansible/ansibleCommandLineInterface.ts
onetocny: Done in 37f5a7a7 — switched the local bindings (inventoryFilePath, playbookPath, sudoUser, dditionalParams) to const.

Extensions/Ansible/Src/Tasks/Ansible/ansibleCommandLineInterface.ts
```diff
@@ -158,29 +175,77 @@ export class ansibleCommandLineInterface extends ansibleInterface {
...
     }
 
+    // Returns the hardened value when the fix is activated, otherwise the legacy
+    // value. The hardening transform is provided as a lazy producer so that the
+    // shellQuote / shellSplit / neutralizeCommandSubstitution helpers are only
+    // ever invoked when the feature flag is on — with the flag off the task
+    // behaves exactly as before and is unaffected by those helpers. Records
+    // fields that contain shell metacharacters so that the telemetry flag can
+    // report the injection surface during rollout.
+    private _applyHardening(fieldName: string, rawValue: string, legacyValue: string, harden: (value: string) => string): string {
```
onetocny: `rawValue` is almost always equal to `legacyValue`. Create new overload of `_applyHardening` that accepts only fieldName, rawValue and harden and pass them to current `_applyHardening` implementation.

onetocny: Done in e49b8b6d — added a 3-argument overload \_applyHardening(fieldName, rawValue, harden)\ that defaults \legacyValue\ to \
awValue\ and delegates to the full implementation. The four call sites where raw == legacy now use it; the host-list site keeps the 4-arg overload since its legacy value is the quoted form.

wawanawna:
🔴 **Remote inventory path is still injectable when the flag is ON**

This happens for `remoteMachine` + `inventoryType=file` + `inventoryFileSource=agentMachine`.

`copyInventoryAndSetPathForAgentAsSource()` sets `this._inventoryPath = '/tmp/' + path.basename(inventoryFile)`. This runs before `super._executeAnsiblePlaybook()`.

Now `_inventoryPath` is set. So the base class skips its `shellQuote` step. Then `_buildPlaybookExecutionCommand()` adds it with no quotes: `'-i ' + this._inventoryPath`.

`path.basename()` does not remove shell characters. So a file named `inv;$(id).ini` becomes `-i /tmp/inv;$(id).ini`. The `;` and `$()` then run on the SSH target. (Confirmed with a local POSIX-shell repro: the `$(id)` payload executes.)

So this path is not protected, even with the flag ON. The comment `// already hardened where it is produced` is not true here. (`playbookPath` is safe, because it is quoted later during build.)

**The same methods have two more unquoted shell sinks.** The cleanup commands use the same raw path: `this._cleanupCmd.push('rm -f ' + remoteInventory)`, and in `copyPlaybookAndSetPathForAgentAsSource()`, `'rm -rf ' + remotePlaybookRoot` (both built from `/tmp/` + `basename(userInput)`). They run later over SSH too. So `rm -f /tmp/inv;$(id).ini` injects as well. Quoting only `_inventoryPath` would leave these open.

**Fix:**
- Read `getBoolFeatureFlag` before the copy methods run.
- `shellQuote` the `/tmp/...` path everywhere it goes into a shell command: the `-i` argument **and** the `rm -f` / `rm -rf` cleanup commands. Keep the raw (unquoted) path for the SFTP copy, which does not use a shell.
- The flags are set in `super`, which runs after the copy. So calling `_applyHardening` in the copy method does nothing (the flags are still `false`).
- Keep the flag-OFF output the same as today.
- Please add a `remoteMachine` + `file` + `agentMachine` test with a file name that has shell characters. This path has no test today. That is why the bug was missed.
- Note: do not just centralize inventory quoting in `_buildPlaybookExecutionCommand`. The file and hostList paths already return quoted values. Quoting again there would double-quote them.

wawanawna:
🔴 **The flag silently corrupts JSON `--extra-vars`**

Look at `shellSplit(args).map(neutralizeCommandSubstitution).join(' ')`.

`shellSplit` removes the outer quotes **and** the inner quotes. So `--extra-vars '{"a":"b"}'` becomes `--extra-vars {a:b}`.

This is worse than a clean error. `ansible-playbook` reads `--extra-vars` as YAML, so `{a:b}` does not fail. It loads as a variable named `a:b` set to null, and the real `a` is gone. Spaced values can also change type (for example `"123"` becomes the number `123`). So the run keeps going, just with the wrong variables.

The problem is inside `shellSplit`. It drops the inner `"..."` after the outer `'...'`. So `neutralize` never sees valid JSON, and re-quoting later cannot bring it back.

JSON extra-vars is common in Ansible. So turning on `AZP_75787_ENABLE_NEW_LOGIC` will silently break real pipelines. The escape logic itself is fine. The tokenizer is the problem.

**Fix:**
- Change the tokenizer.
- It must still keep unquoted `$VAR`. Today `$HOME` and `${HOME}` are kept on purpose.
- It must also handle globs.
- Do not use `shell-quote`'s `parse()`. It is not safe here. With no env it turns `$VAR` into empty text. It also returns glob/operator objects, not strings.
- Note: `shellSplit` lives in the shared `azure-pipelines-tasks-utility-common` package. A change there affects other tasks.
- Please add a round-trip test. Check that flag-ON equals legacy for these inputs: `--extra-vars '{"a":"b"}'`, an arg with `$HOME`, and an arg with a quoted space.

wawanawna:
🟠 **The telemetry cannot show which runs the fix will change**

`_applyHardening` records a field only when the raw value matches `_shellMetaRegex`. That is not the same as "the command changed".

`shellQuote` always adds quotes. So it can change the command even when the regex does not match. Example: a path with a space, like `/opt/my play/site.yml`. The regex does not match it. But the command still changes. (And the old, unquoted command was actually broken.)

So COLLECT can miss the runs that change.

Also, telemetry only publishes when `AZP_75787_ENABLE_COLLECT` is ON. The two flags are separate. So if `NEW_LOGIC` is ON but `COLLECT` is OFF, nothing is sent. You cannot turn on the new escaping and still see which runs it changes, unless you also turn on `COLLECT`.

**Fix:**
- Compute both `legacyValue` and `hardenedValue` every time.
- Use `legacyValue !== hardenedValue` as the signal. That is the true "will change" set.
- Record it when either flag is on.
- Tag the mode (shadow vs live). Then a run with both flags on is not confusing.
- Small note: the code reads only two flags (`AZP_75787_ENABLE_NEW_LOGIC`, `AZP_75787_ENABLE_COLLECT`). It imports the helpers from `azure-pipelines-tasks-utility-common`. If the PR description still mentions an `_LOG` audit mode or vendored/copied helpers, please update it to match the code.

onetocny:
> 🔴 Remote inventory path is still injectable when the flag is ON

Fixed in 01dcfd33.

- The feature flags are now read in the `ansibleCommandLineInterface` **constructor**, so they are resolved before `ansibleRemoteMachineInterface` copies artifacts and sets `_inventoryPath` / pushes cleanup commands. (The idempotent `_readSanitizationFlags` helper was later removed — the constructor read is sufficient.)
- `copyInventoryAndSetPathForAgentAsSource()` now sets `this._inventoryPath = this._applyHardening('inventoryFile', remoteInventory, shellQuote)` and reuses that hardened value in the `rm -f` cleanup command.
- `copyPlaybookAndSetPathForAgentAsSource()` hardens the `rm -rf` cleanup via `_applyHardening('playbookCleanupPath', remotePlaybookRoot, shellQuote)`.
- The raw (unquoted) path is kept for the SFTP/SCP copy, which does not go through a shell.
- Flag-OFF output is unchanged (`_applyHardening` returns the legacy value when both flags are off).
- Added an L0 test `testCommandInjectionHardeningForRemoteAgentSource.ts` covering `remoteMachine` + `file` + `agentMachine` with shell metacharacters in the file name.

onetocny:
> 🔴 The flag silently corrupts JSON `--extra-vars`

Fixed in 01dcfd33.

- Replaced the `shellSplit(args).map(neutralizeCommandSubstitution).join(' ')` tokenizer with a **quote-aware single-pass transform**, `neutralizeAdditionalParameters`, extracted into its own module (`argsSanitization.ts`) so it can be unit-tested in isolation.
- The new transform never re-tokenizes, so inner quotes are preserved: single-quoted content is passed through untouched; `$(...)` / backticks are neutralized even inside double quotes; control operators (`; | & < > ( ) #`) are escaped only when outside quotes; newlines are stripped.
- Unquoted `$VAR` / `${VAR}` and globs are preserved (not escaped), matching the previous intent.
- I did **not** touch `shellSplit` in the shared `azure-pipelines-tasks-utility-common` package (avoids cross-task blast radius) — the fix is local to this task.
- Added round-trip tests confirming flag-ON == legacy for `--extra-vars '{"a":"b"}'`, `$HOME`/`${HOME}`, and quoted-space args, in `argsSanitizationTests.ts` (27 unit cases) plus a scenario test `testExtraVarsJsonPreservedWhenHardeningEnabled.ts`.

onetocny:
> 🟠 The telemetry cannot show which runs the fix will change

Fixed in 01dcfd33.

- `_applyHardeningLegacy` now computes both `legacyValue` and `hardenedValue = harden(rawValue)` and uses `legacyValue !== hardenedValue` as the "will change" signal, replacing the old `_shellMetaRegex` match (which missed e.g. space-only paths that `shellQuote` still changes). `_shellMetaRegex` was removed.
- The field is recorded when **either** flag is on (the hardening helper is only invoked when at least one flag is set, so the flag-OFF path is untouched).
- The telemetry payload now carries a `mode: 'live' | 'shadow'` tag (`live` when `NEW_LOGIC` is on, `shadow` when only `COLLECT` is on) so a both-flags-on run is unambiguous.
- Added a scenario test `testSanitizationTelemetryShadowForAgentMachine.ts` covering the shadow-mode signal.
- Re: the note — the code reads only `AZP_75787_ENABLE_NEW_LOGIC` and `AZP_75787_ENABLE_COLLECT` via `tl.getPipelineFeature`, and imports the escaping helpers from `azure-pipelines-tasks-utility-common` (no vendored copy, no `_LOG`/audit mode). The PR description has been updated to match.

onetocny:
**Rollout flags: switched to a dedicated set (independent of `AZP_75787_*`).**

The hardening was initially gated behind the shared `AZP_75787_*` args-sanitizer flags. That would couple the Ansible rollout to the org-wide sanitizer program — flipping the shared flag elsewhere would also change Ansible behaviour (and vice versa). To stage this fix independently, the task now reads dedicated flags keyed to its own work item (Bug 2457936):

- `AZP_2457936_ENABLE_NEW_LOGIC` — enforce (apply hardening)
- `AZP_2457936_ENABLE_COLLECT` — telemetry / shadow mode

Same code shape and semantics as before (constructor read via `tl.getPipelineFeature`); only the flag names changed. Tests and the PR description are updated accordingly. Version bumped to 0.279.50.

wawanawna (approved):

wawanawna (approved):

PR #1517, opened 2026-08-27
Add ESRP release pipelines for Artifact Engine

manolerazvan:
## Summary
- add separate release pipelines for ArtifactEngine V1 and V2
- reuse the shared publisher, Key Vault, service connection, and `npm-release` variable group
- package through the `npm-packages/packages` artifact contract
- default `dryRun` to `true` and restrict production publication to `master`

## Validation
- both pipeline YAML files parse successfully
- whitespace validation passed
- V1 contract uses Node 20, the `ArtifactEngine` suite, and an `artifact-engine@1.x` guard
- V2 contract uses Node 24, the `ArtifactEngineV2` suite, and an `artifact-engine@2.x` guard

### DIFF
--- a/.pipelines/1es-migration/artifact-engine-v1-release.yml
+++ b/.pipelines/1es-migration/artifact-engine-v1-release.yml
@@ -147,6 +147,7 @@ extends:
               clientid: $(EsrpClientId)
               intent: PackageDistribution
               contenttype: npm
+              productstate: legacy
               contentsource: Folder
               folderlocation: $(Pipeline.Workspace)/npm-packages/packages
               waitforreleasecompletion: true
--- a/.pipelines/1es-migration/artifact-engine-v1-release.yml
+++ b/.pipelines/1es-migration/artifact-engine-v1-release.yml
@@ -1,0 +1,157 @@
+name: ArtifactEngineV1 npm - $(Date:yyyyMMdd)$(Rev:.r)
+appendCommitMessageToRunName: false
+
+trigger: none
+pr: none
+
+parameters:
+- name: dryRun
+  displayName: Build and pack without publishing
+  type: boolean
+  default: true
+
+variables:
+- group: esrp-npm-release
+
+resources:
+  repositories:
+  - repository: 1ESPipelineTemplates
+    type: git
+    name: 1ESPipelineTemplates/1ESPipelineTemplates
+    ref: refs/tags/release
+
+extends:
+  template: v1/1ES.Official.PipelineTemplate.yml@1ESPipelineTemplates
+  parameters:
+    settings:
+      networkIsolationPolicy: Permissive,CFSClean,CFSClean2
+    sdl:
+      sourceAnalysisPool:
+        name: 1ES-ABTT-Shared-Pool
+        image: abtt-windows-2025
+        os: windows
+    customBuildTags:
+    - 1ES-AzureExtensions
+    - ArtifactEngineV1
+    stages:
+    - stage: BuildAndPack
+      displayName: Build, test, and pack ArtifactEngineV1
+      jobs:
+      - job: pack
+        displayName: Build and pack npm package
+        pool:
+          name: 1ES-ABTT-Shared-Pool
+          image: abtt-ubuntu-2404
+          os: linux
+        templateContext:
+          outputs:
+          - output: pipelineArtifact
+            targetPath: $(Build.ArtifactStagingDirectory)/npm-packages
+            artifactName: npm-packages
+        steps:
+        - task: NodeTool@0
+          displayName: Use Node.js 20.x
+          inputs:
+            versionSpec: 20.x
+
+        - task: NpmAuthenticate@0
+          displayName: Authenticate with internal npm feed
+          inputs:
+            workingFile: $(Build.SourcesDirectory)/.npmrc
+
+        - script: npm ci --userconfig $(Build.SourcesDirectory)/.npmrc
+          displayName: Install repository dependencies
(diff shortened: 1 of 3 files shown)


rm-dmiri (approved):

.pipelines/1es-migration/artifact-engine-v1-release.yml
```diff
@@ -0,0 +1,157 @@
...
+          - task: EsrpRelease@12
+            displayName: Publish package to npm via ESRP
+            inputs:
+              connectedservicename: $(EsrpServiceConnection)
+              usemanagedidentity: true
+              keyvaultname: $(EsrpKeyVault)
+              signcertname: $(EsrpSignCert)
+              clientid: $(EsrpClientId)
+              intent: PackageDistribution
+              contenttype: npm
```
nemanjarogic, line 149: Both engines publish the same package name, artifact-engine. The current npm/Azure Artifacts `latest `tag is 2.275.2.

V1 has no `publishConfig.tag`, and ESRP defaults to that tag, so publishing 1.279.1 here would move `latest `back to V1.

Please give V1 an explicit non-latest tag, such as `legacy`, using `productstate `or `publishConfig.tag`, and keep V2 explicitly tagged latest.

.pipelines/1es-migration/artifact-engine-v1-release.yml
```diff
@@ -0,0 +1,157 @@
...
+          displayName: Install repository dependencies
+          workingDirectory: $(Build.SourcesDirectory)
+
+        - script: npx gulp build
+          displayName: Build extensions
+          workingDirectory: $(Build.SourcesDirectory)
+          env:
+            npm_config_userconfig: $(Build.SourcesDirectory)/.npmrc
+
+        - script: npx gulp test --suite ArtifactEngine
```
nemanjarogic, line 72: `--suite ArtifactEngine` does not scope normal tests. The gulp task reads `options.suite` only for -`-e2e` and `--perf`, otherwise it calls `resolveSuitesToRun()`.

I reproduced this command and it ran unrelated BitBucket, ExternalTfs, IISWebAppDeploy, ServiceNow, TeamCity, and V2 suites. Please add real suite selection or invoke explicit ArtifactEngine test globs

nemanjarogic: The same goes for v2 pipeline

.pipelines/1es-migration/artifact-engine-v1-release.yml
```diff
@@ -0,0 +1,157 @@
...
+        templateContext:
+          outputs:
+          - output: pipelineArtifact
+            targetPath: $(Build.ArtifactStagingDirectory)/npm-packages
+            artifactName: npm-packages
+        steps:
+        - task: NodeTool@0
+          displayName: Use Node.js 20.x
+          inputs:
+            versionSpec: 20.x
```
nemanjarogic, line 55: Here are my findings:
- V1’s current build graph requires at least Node 18.20 or Node 20.12.1 due to Mocha 11 and Nock 14. `shelljs@0.10.0` declares Node 18+.
- Root build tooling explicitly requires Node 20+
- Node 10 is EOL and incompatible with the current build graph
- ExternalTFS’s graph contains `azure-pipelines-tasks-azure-arm-rest` -> `@azure/identity/@azure/core-*`, which declare `Node >=20`
- TeamCity has no Node-20-only dependency, but its shelljs dependency declares `Node >=18`

-> If we go with Node 20 or 24, we need to ensure we won't have runtime failures.
 V1 compiles to ES6/CommonJS and type-checks against Node 10 declarations. The compiler’s host Node version does not change that configured output target.

Runtime risk comes from dependencies, because they are shipped as JavaScript without being transpiled by V1.

nemanjarogic: If we are thinking about 20 vs 24, maybe it's worth to go with Node 24 because it's active LTS (but we would need to validate this with CI tests).

.pipelines/1es-migration/artifact-engine-v1-release.yml
manolerazvan: Addressed in d1a83f33. V1 now sets EsrpRelease productstate to legacy, so it cannot move latest back to 1.x. V2 explicitly sets productstate to latest. V1 is retained only to validate the pipeline setup for now and will not be used for publication until its runtime compatibility is reviewed.

.pipelines/1es-migration/artifact-engine-v1-release.yml
manolerazvan: Addressed in d1a83f33. resolveSuitesToRun now honors an explicit --suite value before diff-based selection. Validation results: ArtifactEngine resolved 6 files and completed with 69 passing / 1 pending; ArtifactEngineV2 resolved 8 files and completed with 96 passing / 1 pending. In each run, only the requested Running mocha suite label appeared.

.pipelines/1es-migration/artifact-engine-v1-release.yml
manolerazvan: For now V1 is retained only for pipeline validation and will not be published. Its pipeline continues to build on Node 20 because the current root build graph requires Node 20+, but we will review and validate the V1 consumer-runtime/dependency compatibility separately before any V1 release. V2 remains on Node 24 and its focused suite passed (96 passing / 1 pending). This scope is recorded in d1a83f33.

nemanjarogic (approved):

PR #1530, opened 2026-09-02
Refactor gulp tasks to use async/await and improve TypeScript compilation output handling

ivanduplenskikh:
**Description**:
- Wait for the task and test TypeScript compilation streams before completing `compileNode`.
- Stream dependency-install and `tscBuildTasks` output so TypeScript errors are visible in local and CI builds.
- Remove the now-unused debug-output helper.

**Added unit tests:** (N)

**Attached related issue:** (N)

**Checklist**:

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -3,6 +3,7 @@ const cp = require('node:child_process');
 const fs = require('node:fs');
 const os = require('node:os');
 const path = require('node:path');
+const { finished } = require('node:stream/promises');
 const { parseArgs } = require('node:util');
 
 // build/test script
@@ -93,7 +94,7 @@ gulp.task("compilePS", gulp.series("clean", function () {
     }
 }));
 
-gulp.task("compileNode", gulp.series("compilePS", function (cb) {
+gulp.task("compileNode", gulp.series("compilePS", async function () {
     try {
         // Cache all externals in the download directory.
         // Cache all externals in the download directory.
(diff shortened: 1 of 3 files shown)


gulpfile.js
```diff
@@ -93,7 +94,7 @@ gulp.task("compilePS", gulp.series("clean", function () {
     }
 }));
 
-gulp.task("compileNode", gulp.series("compilePS", function (cb) {
+gulp.task("compileNode", gulp.series("compilePS", async function () {
```
starkmsu, line 97: This removes the `cb` parameter, but the `catch` at line 170 still calls `cb(...)`. Any setup failure will therefore throw `ReferenceError: cb is not defined` and hide the original error. Please throw the wrapped error from this async function instead.

ivanduplenskikh: Fixed

gulpfile.js
```diff
@@ -211,8 +211,8 @@ gulp.task("compileNode", gulp.series("compilePS", function (cb) {
     validateArtifactEngineV2TranspileErrors();
 
     // Generate loc files
-    createResjson(cb);
-    cb();
+    createResjson(() => { });
+    await Promise.all([finished(taskCompilation), finished(testCompilation)]);
```
starkmsu: This can leave `compileNode` hanging indefinitely. Both values are unconsumed `gulp.dest()` transform streams, and `finished(stream)` waits for the readable side as well as the writable side. Since nothing drains the readable side, these promises may never resolve. Please either drain the streams or wait only for writable completion, e.g. `finished(taskCompilation, { readable: false })` (and likewise for `testCompilation`).

gulpfile.js
ivanduplenskikh: Fixed

gulpfile.js
ivanduplenskikh: Good point. Thank you. I've fixed it.

starkmsu (approved):

PR #1531, opened 2026-09-03
Support filesystem-aware artifact matching

manolerazvan:
**Description**: Update ArtifactEngine 1.x and V2 to detect the destination filesystem's case behavior. When the existing `CaseInsensitiveArtifactMatchingFixEnabled` Tasks flag is enabled, matching is case-insensitive on case-insensitive Windows and macOS filesystems. Flag-off behavior, Linux, case-sensitive macOS volumes, and detection failures remain case-sensitive.
 
 ArtifactEngine 1.x remains a separate implementation because it has an independent package lifecycle and is still consumed by the TeamCity and External TFS tasks. 
 
 **Documentation changes required:** N - this is an internal matching correction behind an existing dynamic feature flag.
 
 **Added unit tests:** Y - added coverage for feature/platform gating, case-sensitive and case-insensitive filesystems, mount boundaries, empty and missing destinations, detection caching, and failure fallback.
 
 **Attached related issue:** N
 
 **Checklist**:
 - [x] Version was bumped - ArtifactEngine 1.x to `1.280.0` and ArtifactEngineV2 to `2.280.0`.
 - [x] Checked that applied changes work as expected - ArtifactEngine 1.x: 89 passing; focused ArtifactEngineV2 suites: 43 passing.

### DIFF
--- a/Extensions/ArtifactEngine/Engine/artifactEngine.ts
+++ b/Extensions/ArtifactEngine/Engine/artifactEngine.ts
@@ -157,9 +157,15 @@ export class ArtifactEngine {
     }
 
     private isCaseInsensitiveArtifactMatchingEnabled(destProvider: models.IArtifactProvider): boolean {
-        if (!tl.getPipelineFeature('CaseInsensitiveArtifactMatchingFixEnabled')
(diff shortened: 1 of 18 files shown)


wawanawna (approved):
LGTM

